PR #3, opened 2017-04-11
Added in Not Eligible Page

cavster:


conf/messages
```diff
@@ -9,6 +9,14 @@ hts.core.start.par-3=You must also be a UK resident, or a Crown servant or membe
 hts.core.start.button=Start now
 hts.core.start.elsewhere=Elsewhere on GOV.UK
 
+hts.core.not-eligible.header-page=Help To Save
+hts.core.not-eligible.header=You''re not eligible for Help to Save
+hts.core.not-eligible.par-1=To be eligible for an account you, or your spouse or civil partner, need to be getting one of these benefits:
+hts.core.not-eligible.benefit-1=Working Tax Credit
+hts.core.not-eligible.benefit-2=Universal Credit, and your household earnings need to be at least £120 per week
```
sureshhmrc: instead of using `1` and `2` , i prefer to use meaningful text something like `hts.core.not-eligible.benefit.working.tax.credit` and `hts.core.not-eligible.benefit.universal.credit` .. etc

AndyHWChung: I agree, the messages are easier to reuse if they have proper names

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSave.scala
```diff
@@ -44,6 +44,10 @@ trait HelpToSave extends FrontendController {
 		Future.successful(Ok(uk.gov.hmrc.helptosavefrontend.views.html.register.declaration(user)))
   }
 
+  val notelgibible = Action.async { implicit request ⇒
```
AndyHWChung: camel case?

conf/messages
cavster: For this messages we really just use them in one place...But I will change the names ;)

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSave.scala
cavster: Whoops

PR #16, opened 2017-04-28
User details now retrieved

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/util/package.scala
```diff
@@ -0,0 +1,51 @@
...
+import cats.data.EitherT
+import play.api.libs.json.{JsError, Reads}
+import play.mvc.Http
+import uk.gov.hmrc.play.http.HeaderCarrier
+import uk.gov.hmrc.helptosavefrontend.util.JsErrorOps._
+import scala.concurrent.{ExecutionContext, Future}
+
+package object util {
+
+  type NINO = String
```
AndyHWChung: Having it as a type avoids having to read and write ninos like
```
val nino = Nino("ABC123")
...
val nino2 = nino.id
```
Also for JSON reading and writing having this:
```
case class Person(name: String, nino: NINO)
```
would result in the default `Reads` and `Writes` instances expecting and writing this kind of JSON:
```
{
  "name" : "Andy"
  "nino" : {
     "id": "ABC123"
  }
}
```
instead of
```
{
  "name" : "Andy"
  "nino" : "ABC123"
}
```
Although we do get type safety with having it as a case class.

app/uk/gov/hmrc/helptosavefrontend/connectors/CitizenDetailsConnector.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def getDetails(nino: NINO)(implicit hc: HeaderCarrier, ec: ExecutionContext): Result[CitizenDetailsResponse]
+}
+
+object CitizenDetailsConnector {
+
+  case class Address(line1: String, line2: String, line3: String, line4: String,
+                     line5: String, postcode: String, country: String)
+
+  case class CitizenDetailsResponse(address: Address)
```
AndyHWChung: I put it here because it was specific to the `CitizensDetailsConnector` rather than being used by the rest of the project. This `Address` is specific to the format returned by citizens details - at the moment in `UserInfo` the address is modelled by `List[String]`. I didn't want confusion as to why in some places we were using one data structure and in others another.

PR #20, opened 2017-05-03
play-ui-upgrade-7.2.1: Upgrade

adwelly:


tim-holbrook:
As discussed, this is the security fix on our side for the Redirect Security Alert in 7.2.0 so fine from my perspective; also a requirement for our builds now.

PR #28, opened 2017-05-05
E2e selenium tests

MarcusQA:


AndyHWChung:
What are the advantages to having the selenium tests as part of the main project and not as a separate one?

MarcusQA:
Good question. There are a number of advantages to having the test code in the same repository as the production code and here are some of them:
1. It's clearly visible that a particular version of test code correlates to a particular version of production code if it's in the same repository.
2. Test code is more likely to be updated if it's in the same repository as the production code.
3. It's easier to keep test code and production code in sync if they're in the same repository. If they're separate, a new push of production code might cause tests to fail until those tests are updated; similarly, a new push of test code might cause the tests to fail until the production code is updated.
4. There are instances where the test code might want to call the production code. For example, web API tests (making and receiving HTTP requests) might want to call the same DTOs as the production code for serialising/deserialising JSON. Or the test code may wish to call production code to set up data in the database (unlikely to be applicable in our case because generally we don't persist important data in MDTP).

Hope this makes sense - please let me know if not.

Thanks

Marcus

e2e-Selenium-test/src/test/scala/hts/driver/Browser.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.scalatest.selenium.{Chrome, Firefox}
+import org.openqa.selenium.chrome.ChromeDriver
+import java.util.concurrent.TimeUnit
+
+object Browser extends StartUpTearDown {
+  def getOs = System.getProperty("os.name")
+
+  lazy val systemProperties = System.getProperties
+  lazy val isMac: Boolean = getOs.startsWith("Mac")
+  lazy val isLinux: Boolean = getOs.startsWith("Linux")
```
cavster: What about a windows ;)

cavster: Also why are these values lazy?

e2e-Selenium-test/src/test/scala/hts/driver/Browser.scala
```diff
@@ -0,0 +1,129 @@
...
+  lazy val linuxArch = systemProperties.getProperty("os.arch")
+  override val driver: WebDriver = null
+  var jsStatus = true
+
+  object PhantomJSDriverObject {
+    def apply(capabilities: DesiredCapabilities) = new PhantomJSDriver(capabilities)
+  }
+
+  def javascriptEnabled: Boolean = {
+    if (!StringUtils.isEmpty(systemProperties.getProperty("javascriptEnabled")))
```
cavster: Don't need  the is and else marcus :)

e2e-Selenium-test/src/test/scala/hts/driver/Browser.scala
```diff
@@ -0,0 +1,129 @@
...
+      true
+    else
+      false
+  }
+
+  def createGeckoDriver(): WebDriver = {
+    if (StringUtils.isEmpty(systemProperties.getProperty("webdriver.gecko.driver"))) {
```
cavster: Pattern match it ;)

AndyHWChung:
Moved to https://github.com/hmrc/help-to-save-frontend/pull/29

PR #29, opened 2017-05-09
Added selenium tests

AndyHWChung:


AndyHWChung:
What are the advantages to having the selenium tests as part of the main project and not as a separate one?

Reply from Marcus:
_Good question. There are a number of advantages to having the test code in the same repository as the production code and here are some of them:

It's clearly visible that a particular version of test code correlates to a particular version of production code if the test code is in the same repository.
Test code is more likely to be updated if it's in the same repository as the production code.
It's easier to keep test code and production code in sync if they're in the same repository. If they're separate, a new push of production code might cause tests to fail until those tests are updated; similarly, a new push of test code might cause the tests to fail until the production code is updated.
There are instances where the test code might want to call the production code. For example, web API tests (making and receiving HTTP requests) might want to call the same DTOs as the production code for serialising/deserialising JSON. Or the test code may wish to call production code to set up data in the database (unlikely to be applicable in our case because generally we don't persist important data in MDTP).
Of course, we must make sure we don't allow any tests to run in production!__

PR #33, opened 2017-05-11
HTS-107: Implement new play-auth 

sureshhmrc:


test/uk/gov/hmrc/helptosavefrontend/services/userinfo/HelpToSaveServiceSpec.scala
```diff
@@ -114,13 +92,14 @@ class UserInfoServiceSpec extends UnitSpec with WithFakeApplication with MockFac
         ))
 
         // test if user details does not have the surname
```
cavster: Probable don't need these comments ;)

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveController.scala
```diff
@@ -16,24 +16,69 @@
...
+
+  def authorisedForHts(body: => Future[Result])(implicit request: Request[AnyContent]): Future[Result] = {
+    authorised(AuthProviders(GovernmentGateway)) {
+      body
+    } recoverWith {
+      handleFailure
+    }
+  }
+
+  def handleFailure(implicit request: Request[_], messagesApi: MessagesApi): PartialFunction[Throwable, Future[Result]] =
```
AndyHWChung: the type signature could just be `Throwable => Result` - you don't need a partial function in this case since you are defining the function for all `Throwables` anyway

AndyHWChung: Also, you aren't actually making any calls to any `Futures` inside your function

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveController.scala
```diff
@@ -16,24 +16,69 @@
...
+        val nino =
+          allEnrols.enrolments.find(enrol ⇒ enrol.key == "HMRC-NI")
+            .getOrElse(throw HelpToSaveException("No HMRC-NI enrolment found for logged in user"))
+            .getIdentifier("NINO")
+            .getOrElse(throw HelpToSaveException("No NINO found for logged in user"))
+            .value
+
+        body(userUri.get, nino)
+
+    } recoverWith {
```
AndyHWChung: Does the `play-auth` library throw exceptions? Maybe instead of try-catch wrap the parts you expect to fail in a `Try` - it is then more apparent in the types which parts throw exceptions - also we shouldn't be throwing exceptions in our own code.

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveController.scala
```diff
@@ -16,24 +16,69 @@
...
-  protected type AsyncPlayUserRequest = AuthContext => Request[AnyContent] => Future[Result]
+class HelpToSaveController extends FrontendController with AuthorisedFunctions with Redirects {
 
   override def authConnector: AuthConnector = FrontendAuthConnector
 
-  def AuthorisedHtsUserAction(body: AsyncPlayUserRequest): Action[AnyContent] =
-    AuthorisedFor(HtsRegime, HtsCompositePageVisibilityPredicate).async(body)
-}
\ No newline at end of file
+  val config: Configuration = Play.current.configuration
```
AndyHWChung: `Play.current` is deprecated - use depency injection

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveController.scala
```diff
@@ -16,24 +16,69 @@
...
+  val env: Environment = Environment(Play.current.path, Play.current.classloader, Play.current.mode)
+
+  case class HelpToSaveException(message: String) extends RuntimeException(message)
+
+  def authorisedForHts(body: (String, String) => Future[Result])(implicit request: Request[AnyContent]): Future[Result] = {
```
AndyHWChung: It isn't clear what `(String,String)` refers to - use a case class?

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveController.scala
```diff
@@ -16,24 +16,69 @@
...
+    def authorisedWithNino = authorised(Enrolment("HMRC-NI").withConfidenceLevel(ConfidenceLevel.L200))
+
+    val compositeRetrieval = userDetailsUri and allEnrolments
+
+    authorisedWithNino.retrieve(compositeRetrieval) {
+      case userUri ~ allEnrols =>
+
+        val nino =
+          allEnrols.enrolments.find(enrol ⇒ enrol.key == "HMRC-NI")
+            .getOrElse(throw HelpToSaveException("No HMRC-NI enrolment found for logged in user"))
```
AndyHWChung: Errors should be encoded in the types we use - an `Option` or `Either` would be better here

test/uk/gov/hmrc/helptosavefrontend/controllers/IvControllerSpec.scala
```diff
@@ -16,79 +16,56 @@
...
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
 
 class IvControllerSpec extends UnitSpec with WithFakeApplication with MockFactory {
 
   val ivConnector: IvConnector = mock[IvConnector]
 
-  val jId = randomUUID().toString
+  private val jId = randomUUID().toString
```
AndyHWChung: Is this `val` needed  - can you not just do `val journeyId = JounreyId(randomUUID().toString)`

test/uk/gov/hmrc/helptosavefrontend/controllers/MockAuthSupport.scala
```diff
@@ -0,0 +1,64 @@
...
+           "key": "HMRC-NI",
+           "identifiers": [{"key":"NINO","value":"WM123456C"}],
+           "state": "Activated",
+           "confidenceLevel": 200
+          }
+       ]}
+    """.stripMargin
+  ))
+
+  val mockPlayAuth = new PlayAuthConnector {
```
AndyHWChung: Can you not use scalamock for this? This would be more consistent with the rest of the test code

AndyHWChung: Or do you just always want it to return a successful response to a POST?

sureshhmrc:
@adwelly  and @Jackie1988  could you please take a look at the comments and merge PR ?

PR #53, opened 2017-05-23
Rename "register/declaration" to "register/user-details"

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/config/frontendAppConfig.scala
```diff
@@ -35,7 +35,7 @@ object FrontendAppConfig extends AppConfig with ServicesConfig {
   private val contactHost = configuration.getString(s"contact-frontend.host").getOrElse("")
   private val contactFormServiceIdentifier = "MyService"
 
-  val HtsDeclarationUrl = getConfString("help-to-save-declaration.url", "")
+  val HtsUserInfoUrl = getConfString("help-to-save-user-details.url", "")
```
AndyHWChung: We don't need to read this from config - we can use `uk.gov.hmrc.helptosavefrontend.controllers.routes.RegisterController.declaration().url` whenever we need to reference that url in the code

AndyHWChung:
LGTM

PR #57, opened 2017-05-25
HTS-210 : Make sure all the /register endpoints are protected with auth + CL200

sureshhmrc:


test/uk/gov/hmrc/helptosavefrontend/controllers/RegisterControllerSpec.scala
```diff
@@ -48,7 +47,7 @@ class RegisterControllerSpec extends TestSupport with ScalaFutures {
   private val enrolment = Enrolment("HMRC-NI", Seq(EnrolmentIdentifier("NINO", nino)), "activated", ConfidenceLevel.L200)
   private val enrolments = Enrolments(Set(enrolment))
 
-  val mockAuthConnector = mock[PlayAuthConnector]
+  private val mockAuthConnector = mock[PlayAuthConnector]
```
AndyHWChung: `private` for unit tests are probably not needed - are we ever going to extend this spec or try to access its members from outside? But having it there doesn't really hurt either - it just clutters up the code a bit

PR #58, opened 2017-05-25
Added new Selenium scenarios

MarcusQA:


selenium-system-test/src/test/resources/features/security/Applicant CANNOT proceed without going through security.feature
```diff
@@ -1,11 +1,23 @@
 Feature: Applicant CANNOT proceed without going through security
 
-  Scenario: Applicant CANNOT proceed with application if they have NOT passed 2SV
-    Given an applicant has a confidence level of 200
-    But their credential strength is weak
-    Then they are forced into going through 2SV before being able to proceed with their HtS application
+  Scenario: Applicant CANNOT view the user details page without logging in
+    Given an applicant has NOT logged in
+    When they try to view the user details page
+    Then they are forced to log in
```
adwelly: 'forced to log in' ? 
I wonder if this gets to the nub of what actually happens - e.g. they see the login page and are invited to login

selenium-system-test/src/test/scala/hts/driver/Driver.scala
```diff
@@ -31,6 +30,7 @@ object Driver {
   private val systemProperties = System.getProperties
 
   def newWebDriver(): Either[String, WebDriver] = {
+    println("Creating new driver\n\n")
```
adwelly: Do yuo still need this?

selenium-system-test/src/test/resources/features/security/Applicant CANNOT proceed without going through security.feature
MarcusQA: I'm trying to avoid referring to UI elements in the Gherkin scenarios and, instead, would like to just present the hard business rules. How about "Then they are prompted to log in"?

selenium-system-test/src/test/scala/hts/driver/Driver.scala
MarcusQA: No - please feel free to remove :)

adwelly:
QA believes all selenium tests will run given an appropriate browser version.

PR #63, opened 2017-05-31
HTS-220 : fix "You're not eligible" url

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -88,12 +88,14 @@ class RegisterController @Inject()(val messagesApi: MessagesApi,
       )
   }
 
-
   def accessDenied: Action[AnyContent] = Action.async {
     implicit request ⇒
       Future.successful(Ok(views.html.access_denied()))
   }
 
+  val notEligible: Action[AnyContent] = Action.async { implicit request ⇒
```
AndyHWChung: Could you not just write:
```scala
val notEligible: Action[AnyContent] = Actions { implicit request => Ok(views.html.core.not_eligible() } 
```

PR #65, opened 2017-06-01
Implemented scenario for user details

MarcusQA:


selenium-system-test/src/test/scala/hts/steps/UserDetailsSteps.scala
```diff
@@ -0,0 +1,65 @@
...
+
+class UserDetailsSteps extends Steps{
+
+  var name: Option[String] = None
+  var nino: Option[String] = None
+  var dateOfBirth: Option[String] = None
+  var email: Option[String] = None
+
+  Given("""^an applicant has the following details:$"""){ (applicantDetails:DataTable) =>
+    val expectedData = applicantDetails.asMaps(classOf[String], classOf[String])
```
AndyHWChung: this returns a java `List` instead of a scala `List`. Working with a scala `List` will allow you to do a `foreach` instead of the `for` loop you have below

selenium-system-test/src/test/scala/hts/steps/UserDetailsSteps.scala
```diff
@@ -0,0 +1,65 @@
...
+      val row = expectedData.get(i).asScala.seq
+      val field = Option(row("field")).getOrElse("")
+      val value = Option(row("value"))
+
+      field match {
+        case "name" => name = value
+        case "NINO" => nino = value
+        case "date of birth" => dateOfBirth = value
+        case "email address" => email = value
+      }
```
adwelly: So I think there may be a better way of handling field here along the line of

val field = row("field")

and having a case _ => in the match to catch the null case.

If you would rather not, I think you need to look into handling the situation where field = "" in any case.

selenium-system-test/src/test/scala/hts/steps/UserDetailsSteps.scala
```diff
@@ -0,0 +1,65 @@
...
+    AuthorityWizardPage.goToPage()
+    AuthorityWizardPage.setRedirect(Configuration.host + "/help-to-save/register/user-details")
+    AuthorityWizardPage.setCredentialStrength("strong")
+    AuthorityWizardPage.setConfidenceLevel(200)
+    AuthorityWizardPage.setNino(nino.getOrElse(""))
+    AuthorityWizardPage.submit()
+  }
+
+  Then("""^they see their details$"""){ () =>
+    Page.getPageContent() should include("Name: " + name.getOrElse(""))
```
AndyHWChung: Isn't the `None` cases here error cases? Should the test not fail at this point if they are not present?

PR #85, opened 2017-06-12
Addeed NS&I authorization

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -56,11 +57,23 @@ class NSIConnectorImpl extends NSIConnector with ServicesConfig {
   val nsiUrlEnd: String = getString("microservice.services.nsi.url")
   val url = s"$nsiUrl/$nsiUrlEnd"
 
+  val authorisationHeaderKeys = getString("microservice.services.nsi.authorization.header-key")
```
sureshhmrc: is it ```Keys``` or just a single ```Key``` ?

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -56,11 +57,23 @@ class NSIConnectorImpl extends NSIConnector with ServicesConfig {
...
+
+  val authorizationEncoding = getString("microservice.services.nsi.authorization.encoding")
```
sureshhmrc: we can use short names like ```encoding``` ?

conf/application.conf
```diff
@@ -165,6 +165,14 @@ microservice {
         base64KeystorePassword = ""
         certificate.pem = ""
       }
+      authorization {
+        # the key of the entry in the authorization header in the request going to NS&I -
+        # set to 'Authorization1` for local devleopment
+        header-key = "Authorization1"
+        user       = "user"
+        password   = "password"
+        encoding   = "UTF-8"
```
sureshhmrc: is encoding different for each env ?  if not we can get this off from here

test/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnectorImplSpec.scala
```diff
@@ -48,14 +50,25 @@ class NSIConnectorImplSpec extends UnitSpec with WithFakeApplication with MockFa
...
+
+  val authorisationHeaderKeys = config.getString("microservice.services.nsi.authorization.header-key")
+  val authorizationEncoding = config.getString("microservice.services.nsi.authorization.encoding")
+
+  val authorisationDetails = {
+    val user = config.getString("microservice.services.nsi.authorization.user")
+    val password = config.getString("microservice.services.nsi.authorization.password")
+    val encoded = Base64.getEncoder.encode(s"$user:$password".getBytes)
+    s"Basic: ${new String(encoded, authorizationEncoding)}"
+  }
```
sureshhmrc: this bit is same as above, we can remove the duplication

conf/application.conf
AndyHWChung: I made it into a config variable just in case it isn't UTF-8 for production - I can code the stub to always accept UTF-8 but we don't have control over what NS&I will take

test/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnectorImplSpec.scala
AndyHWChung: not sure what you mean here...

PR #90, opened 2017-06-16
Hts 223

Jackie1988:


selenium-system-test/src/test/scala/hts/steps/EligibilitySteps.scala
```diff
@@ -37,11 +38,11 @@ class EligibilitySteps extends Steps{
   }
 
   Then("""^they see that they are eligible for Help to Save$""") { () =>
-    Page.getPageContent() should include("Check and confirms your details")
+    Page.getPageContent() should include("Check and confirm your details")
   }
 
   Given("""^an applicant is NOT in receipt of working tax credit$""") { () =>
-    nino = Option("JA553215D")
+    nino = Some(generateIllegibleNINO)
```
MarcusQA: Do you mean generateIneligibleNINO?

Jackie1988: Yes you're right

PR #92, opened 2017-06-19
HTS-187: Fix the regex for coverageExcludedPackages(WIP)

sureshhmrc:


test/uk/gov/hmrc/helptosavefrontend/connectors/IvConnectorSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  "The IvConnectorImpl" when {
+
+    "getting Journey Status" should {
+
+      "handle successful response" in new TestApparatus {
+
+        val httpResponse = HttpResponse.apply(200, Some(Json.parse("""{"result": "Success"}""")))
```
AndyHWChung: do you need the explicit `.apply` here?

test/uk/gov/hmrc/helptosavefrontend/connectors/IvConnectorSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+        mockHttpResponse((httpResponse))
+
+        val result = ivConnector.getJourneyStatus(journeyId)
+
+        result.futureValue should be(Some(Success))
+      }
+
+      "handle unexpected non-successful response" in new TestApparatus {
+
+        val httpResponse = HttpResponse.apply(600)
```
AndyHWChung: and the `.apply` here as well?

test/uk/gov/hmrc/helptosavefrontend/services/HelpToSaveServiceSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+      val userDetailsURI = "/dummy/user/details/uri"
+      val nino = "WM123456C"
+
+      "return a successful response" in {
+
+        val user = models.randomUserInfo()
+
+        val eligResult = EligibilityResult(Some(user))
+
+        val eligStatus: Result[EligibilityResult] = EitherT[Future, String, EligibilityResult](Right(eligResult))
```
AndyHWChung: You can also write `EitherT[Future, String, EligibilityResult].pure(eligrResult)`

project/MicroService.scala
```diff
@@ -26,8 +26,8 @@ trait MicroService {
     import scoverage.ScoverageKeys
     Seq(
       // Semicolon-separated list of regexs matching classes to exclude
-      ScoverageKeys.coverageExcludedPackages := "<empty>;Reverse.*;model.*;config.*;.*(AuthService|BuildInfo|Routes).*",
-      ScoverageKeys.coverageMinimum := 60,
+      ScoverageKeys.coverageExcludedPackages := "<empty>;Reverse.*;.*(config|views.*|models.*);.*(AuthService|BuildInfo|Routes).*",
```
AndyHWChung: Do we want to exclude all of the `models` package? There is stuff in `NSIUserInfo` which should be tested but also it probably isn't so important to include it in the coverage statistics. The alternative is to have:
```scala
// $COVERAGE-OFF
...
// $COVERAGE-ON$
```
in all the things in `models` which we don't want, but this would be quite annoying

test/uk/gov/hmrc/helptosavefrontend/connectors/IvConnectorSpec.scala
sureshhmrc: yes, its not a case class

test/uk/gov/hmrc/helptosavefrontend/connectors/IvConnectorSpec.scala
sureshhmrc: yes,

AndyHWChung:
LGTM

PR #96, opened 2017-06-20
HTS-234 :  Improve the code coverage by adding tests

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/config/frontendAppConfig.scala
```diff
@@ -49,6 +50,19 @@ object FrontendAppConfig extends AppConfig with ServicesConfig {
       s"&confidenceLevel=200")
       .toString
 
+  val authHeaderKey = getString("microservice.services.nsi.authorization.header-key")
```
AndyHWChung: This and `authDetails` are particular to NS&I - the name is a bit misleading as we do other authorisation stuff elsewhere in the code

PR #100, opened 2017-06-23
OAuth authorisation code is now obtained and passed onto backend

AndyHWChung:


sureshhmrc:
LGTM , one concern as discussed is try and see if we can move all this code to backend to make the design simpler

PR #109, opened 2017-07-10
HTS-287 : Initial HTML page for showing up missing info errors to the user

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -74,20 +75,17 @@ class RegisterController @Inject()(val messagesApi: MessagesApi,
...
-                  case _ ⇒
-                    Logger.error(s"Could not retrieve logged in user's NINO or userDetails")
-                    SeeOther(routes.RegisterController.accessDenied().url)
+                }, { success ⇒
+                  val nino = success._1
+                  val eligibility = success._2
+                  eligibility.result.fold(
+                    infos ⇒ {
+                      Logger.error(s"user $nino has missing information: ${infos.missingInfo.mkString(",")}")
+                      Ok(views.html.register.missing_user_info(infos.missingInfo, personalAccountUrl))
```
AndyHWChung: What details can you actually change via pertax-frontend?

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -74,20 +75,15 @@ class RegisterController @Inject()(val messagesApi: MessagesApi,
...
-                      }, {
-                        case Some(info) ⇒ Ok(views.html.register.confirm_details(info))
-                        case _ ⇒ SeeOther(routes.RegisterController.notEligible().url)
-                      })
-
-                  case _ ⇒
-                    Logger.error(s"Could not retrieve logged in user's NINO or userDetails")
-                    SeeOther(routes.RegisterController.accessDenied().url)
+                }, { case (nino, eligibility) ⇒
+                  eligibility.result.fold(
```
sureshhmrc: agreed, wanted to get rid of ```case _ => ``` part , but anyway used case now , looks better

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
sureshhmrc: I can see only address can be updated in the /personal-account page ...

AndyHWChung:
Since pertax-frontend can only change the address do we need to only direct them to there if the address is missing?

sureshhmrc:
in general its nice to provide the personal account link and ask user to ```go and update the address and then continue help-to-save account```  but either way i do not have a strong opinion on how it should look like... its upto UX guys

AndyHWChung:
I guess the code lays down the framework code needed to direct the user to the right place, but is there somewhere where they can change the reset of the details? Probably not, but just wondering how much this has been investigated.

AndyHWChung:
The code looks fine - I'll leave it to Marcus to comment on what the redirects should really be

PR #114, opened 2017-07-20
replaced java schema validation library with scala library

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -50,7 +45,8 @@ import scala.concurrent.{ExecutionContext, Future}
 @Singleton
 class RegisterController @Inject()(val messagesApi: MessagesApi,
                                    helpToSaveService: HelpToSaveService,
-                                   sessionCacheConnector: SessionCacheConnector)(implicit app: Application, ec: ExecutionContext)
+                                   sessionCacheConnector: SessionCacheConnector,
+                                   jsonSchemaValidationService: JSONSchemaValidationService)(implicit app: Application, ec: ExecutionContext)
```
adwelly: Could we inject the app rather than using the implicit one?

AndyHWChung: It is being injected, it's just being made available implicitly inside the controller which  is unecessary - taking it out

PR #117, opened 2017-07-24
Removed validation from NSIUserInfo

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/models/NSIUserInfo.scala
```diff
@@ -40,10 +40,6 @@ case class NSIUserInfo (forename: String,
 
 object NSIUserInfo {
 
-  private implicit class StringOps(val s: String) {
```
sureshhmrc: several unused imports

AndyHWChung: whoops, I forgot - taken them out now

PR #123, opened 2017-07-27
HTS-360 : Implement "Sign out" feature for Help to Save and trigger exit survey

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveAuth.scala
```diff
@@ -36,34 +36,31 @@ class HelpToSaveAuth(app: Application) extends FrontendController with Authorise
 
   override def env: Environment = Environment(app.path, app.classloader, app.mode)
 
+  private type HtsAction = Request[AnyContent] ⇒ HtsContext ⇒ Future[Result]
 
-  private type HtsAction = Request[AnyContent] => Future[Result]
-  private type HtsActionWithEnrolments = Request[AnyContent] => Option[NINO] => Future[Result]
-
-  def authorisedForHtsWithEnrolments(action: HtsActionWithEnrolments): Action[AnyContent] = {
+  def authorisedForHtsWithEnrolments(action: HtsAction): Action[AnyContent] =
```
AndyHWChung: this could almost be renamed to `authorisedForHtsWithNino` since the only difference with `authorisedForHts` is that it returns a NINO if possible

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveAuth.scala
```diff
@@ -73,24 +70,28 @@ class HelpToSaveAuth(app: Application) extends FrontendController with Authorise
...
       authorised(AuthWithConfidence) {
-        action(request)
+        action(request)(HtsContext(None, isAuthorised = true))
       }.recover {
         case e ⇒ handleFailure(e)
       }
     }
   }
 
+  def isAuthorised(action: HtsAction): Action[AnyContent] = {
```
AndyHWChung: `isAuthorised` is a bit misleading - it sounds like it should be returning a boolean, but what it's actually doing is performing the action whether or not the user is logged in. Perhaps naming it `unprotected` or something along those lines would be more descriptive and makes it clear that the action will still occur if the user is not logged in

app/uk/gov/hmrc/helptosavefrontend/controllers/StartPagesController.scala
```diff
@@ -19,26 +19,32 @@ package uk.gov.hmrc.helptosavefrontend.controllers
...
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc._
 import uk.gov.hmrc.helptosavefrontend.views
-import uk.gov.hmrc.play.frontend.controller.FrontendController
 
 import scala.concurrent.Future
 
 @Singleton
-class StartPagesController @Inject()(val messagesApi: MessagesApi) extends FrontendController with I18nSupport {
+class StartPagesController @Inject()(implicit val app: Application, val messagesApi: MessagesApi) extends HelpToSaveAuth(app) with I18nSupport {
```
AndyHWChung: do we need the `val` in front of `app` here?

app/uk/gov/hmrc/helptosavefrontend/controllers/HelpToSaveAuth.scala
sureshhmrc: cant think of a better name, lets stick to `unprotected`

AndyHWChung:
LGTM

PR #129, opened 2017-07-31
Hts 1 - refactoring and removing old intro pages

jamesbirrellgray:


sureshhmrc:
specs need to be updated too

PR #145, opened 2017-08-03
HTS-392 : Fix redirection Url

sureshhmrc:


conf/application.conf
```diff
@@ -239,7 +239,7 @@ microservice {
       port = 7001
     }
 
-    help-to-save-confirm-details {
+    help-to-save-check-eligibility {
```
AndyHWChung: should this even be under the microservices section?

AndyHWChung:
LGTM

PR #152, opened 2017-08-09
removed /register/nsi endpoint

AndyHWChung:


AndyHWChung:
tested locally - users can now lo longer put in a url into their browser which will proceed to create an account with NS&I

PR #169, opened 2017-08-22
Hts 416

Jackie1988:


app/uk/gov/hmrc/helptosavefrontend/views/register/update_email_address.scala.html
```diff
@@ -14,15 +14,31 @@
  * limitations under the License.
  *@
 
+@import uk.gov.hmrc.helptosavefrontend.views.html.helpers
 @import uk.gov.hmrc.helptosavefrontend.models.HtsContext
+@import uk.gov.hmrc.helptosavefrontend.forms.UpdateEmail
+@import uk.gov.hmrc.helptosavefrontend.util.FormHelpers
 
-@()(implicit htsContext: HtsContext, request: Request[_], messages: Messages)
+@(emailAddress: String, form: Option[Form[UpdateEmail]])(implicit htsContext: HtsContext, request: Request[_], messages: Messages)
```
AndyHWChung: is there any case where `form` would be `None`?

app/uk/gov/hmrc/helptosavefrontend/util/FormHelpers.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package uk.gov.hmrc.helptosavefrontend.util
+
+import play.api.data.Form
+
+/**
+  * Created by andy on 17/08/2017.
+  */
+object FormHelpers {
```
AndyHWChung: This looks like it belongs in the views - probably under `helpers`

app/uk/gov/hmrc/helptosavefrontend/util/FormHelpers.scala
```diff
@@ -0,0 +1,36 @@
...
+  def getErrorByKey[A](form: Option[Form[A]], errorKey: String) = {
```
AndyHWChung: The return type is `String` but it makes more sense if it was `Option[String]`. There's an easier implementation:
```scala
def getErrorByKey[A](form: Option[Form[A]], errorKey: String): Option[String] =
  form.flatMap(_.error(errorKey).map(_.message))
```

app/uk/gov/hmrc/helptosavefrontend/controllers/UpdateEmailAddressController.scala
```diff
@@ -44,9 +49,29 @@ class UpdateEmailAddressController @Inject()(val sessionCacheConnector: SessionC
...
+            })
           }
         }
   }(redirectOnLoginURL = FrontendAppConfig.checkEligibilityUrl)
 
-}
+  def onSubmit(): Action[AnyContent] = authorisedForHtsWithInfo {
+    implicit request =>
+      implicit htsContext ⇒
+        UpdateEmailForm.verifyEmailForm.bindFromRequest().fold(
```
AndyHWChung: should probably do the `checkIfAlreadyEnrollled` and `checkSession` business here. This'd protect this endpoint. Also, it'll give you access to a non-optional NINO to use below

app/uk/gov/hmrc/helptosavefrontend/controllers/UpdateEmailAddressController.scala
```diff
@@ -44,9 +49,29 @@ class UpdateEmailAddressController @Inject()(val sessionCacheConnector: SessionC
...
+          formWithErrors => {
+            Future.successful(BadRequest(views.html.register.update_email_address("errors", Some(formWithErrors))))
+          },
+          (details: UpdateEmail) => {
+           emailVerificationConnector.verifyEmail(htsContext.nino.getOrElse(""), details.email).map {
+             case Right(x) ⇒ Ok(views.html.register.check_your_email())
+             case Left(AlreadyVerified()) ⇒ Ok(views.html.register.email_verify_error("hts.email-verification.email-verify-error.already-verified.content"))
```
AndyHWChung: you can pass in the `EmailVerificationError` to the view and it can decide what message to get and display rather than passing in the message through the controller

app/uk/gov/hmrc/helptosavefrontend/util/EmailVerificationParams.scala
```diff
@@ -0,0 +1,70 @@
...
+    DataEncrypter.decrypt(base64) match {
+      case Left(_) ⇒ None
+      case Right(decrypted) ⇒
+        val params = decrypted.split(":")
+        if (params.length == 2) Some(EmailVerificationParams(params(0), params(1))) else None
+    }
+  }
+}
+
+object DataEncrypter {
```
AndyHWChung: We swapped out this in the backend to use the hmrc crypto library:

https://github.com/hmrc/help-to-save/blob/master/project/MicroServiceBuild.scala#L24
https://github.com/hmrc/help-to-save/blob/master/app/uk/gov/hmrc/helptosave/util/Crypto.scala

You might consider doing the same here

test/uk/gov/hmrc/helptosavefrontend/controllers/RegisterControllerSpec.scala
```diff
@@ -94,6 +96,45 @@ class RegisterControllerSpec extends TestSupport with EnrolmentAndEligibilityChe
...
+        "and the user has clicked on the verify email link sent to them by the email verification service " in {
+        val testEmail = "email@gmail.com"
+        val theNino = "AE1234XXX"
+        inSequence{
+          mockPlayAuthWithRetrievals(AuthWithConfidence)(userDetailsURIWithEnrolments)
+          mockEnrolmentCheck(nino)(Right(EnrolmentStatus.NotEnrolled))
+          mockSessionCacheConnectorGet(Right(Some(HTSSession(Some(validNSIUserInfo.copy (nino = theNino)), None))))
+        }
+        val params = EmailVerificationParams(theNino, testEmail)
+        val result = doRequestWithQueryParam(params.encode().replaceAll("%2b", "+"))
```
AndyHWChung: would a URL-decode here work to catch all URL-encoded characters? (there's `java.net.URLDecoder` you could use)

test/uk/gov/hmrc/helptosavefrontend/views/register/ViewBehavioursSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import org.jsoup.nodes.Document
+import play.api.i18n.MessagesApi
+import play.api.test.FakeRequest
+import uk.gov.hmrc.helptosavefrontend.TestSupport
+
+class ViewBehavioursSpec extends TestSupport {
+
+  lazy val injector = fakeApplication.injector
+  lazy val request = FakeRequest()
```
AndyHWChung: this one doesn't need to be lazy

test/uk/gov/hmrc/helptosavefrontend/views/register/ViewBehavioursSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+  def messagesApi = injector.instanceOf[MessagesApi]
```
AndyHWChung: this one could be a lazy val - doesn't need to be evaluated every time it's called

AndyHWChung:
@jamesbirrellgray there's a bunch of frontend stuff you probably want to look at as well

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -44,12 +44,26 @@ class RegisterController @Inject()(val messagesApi: MessagesApi,
...
     implicit request ⇒
       implicit htsContext ⇒
         checkIfAlreadyEnrolled { _ ⇒
-          checkIfDoneEligibilityChecks { case (nsiUserInfo, _) ⇒
-            Ok(views.html.register.confirm_details(nsiUserInfo))
+          checkIfDoneEligibilityChecks { case (nsiUserInfo, _) ⇒ {
+            emailVerificationParams match {
+              case None ⇒ Ok(views.html.register.confirm_details(nsiUserInfo))
+              case Some(p) ⇒ {
+                val optionalParams = EmailVerificationParams.decode(p)
```
Jackie1988: OK, we will do that.

Jackie1988: done.

app/uk/gov/hmrc/helptosavefrontend/util/EmailVerificationParams.scala
```diff
@@ -0,0 +1,70 @@
...
+import javax.crypto.{Cipher, SecretKey, SecretKeyFactory}
+import javax.crypto.spec.DESKeySpec
+
+import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig._
+
+import scala.util.control.NonFatal
+
+case class EmailVerificationParams(nino: String, email: String) {
+  def encode(): String = {
+    val input = nino + ":" + email
```
AndyHWChung: if we're going to be really anal about this it's safer to use a character like `§` to separate here and then split using that character when decoding since technically speaking emails may have colons in them. But the chance of someone having a colon in their email is probably remote enough for this not to matter too much.

app/uk/gov/hmrc/helptosavefrontend/util/EmailVerificationParams.scala
Jackie1988: Done.

test/uk/gov/hmrc/helptosavefrontend/controllers/RegisterControllerSpec.scala
Jackie1988: ok done

test/uk/gov/hmrc/helptosavefrontend/views/register/ViewBehavioursSpec.scala
Jackie1988: I have made the request un-lazy :)

test/uk/gov/hmrc/helptosavefrontend/views/register/ViewBehavioursSpec.scala
Jackie1988: ok updated

app/uk/gov/hmrc/helptosavefrontend/connectors/EmailVerificationConnector.scala
```diff
@@ -43,34 +43,33 @@ trait EmailVerificationConnector {
...
           logger.info("[EmailVerification] - Email verification successfully triggered")
           Right(())
         case BAD_REQUEST ⇒
           logger.warn("[EmailVerification] - Bad Request from email verification service")
-          Left(RequestNotValidError(nino))
+          Left(RequestNotValidError())
         case CONFLICT ⇒
-          logger.info("[EmailVerification] - Email already verified")
-          Left(AlreadyVerified(nino, newEmail))
+          logger.info(s"[EmailVerification] - Email: $newEmail already verified")
```
Jackie1988: Tbh no we don't as the new email address either going to be saved in the email collection or it isn't, theres no need to log unsaved email addresses as the user will be notified of the reason and we will be logging the nino in the backend for all unsuccessful cases so we can follow unhappy user journeys in Kibana and figure out whats going wrong for them.

AndyHWChung: I was coming more from a security/sensitive data angle - are emails considered sensitive personal data?

adwelly: I think we do have to treat both emails and ninos as sensitive data - but also that we should be logging at this point. Dumping stuff into the audit logs just because it is sensitive is frowned upon, so we have a bit of an issue here. Could we make a call to the service to record the situation?

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -44,12 +44,26 @@ class RegisterController @Inject()(val messagesApi: MessagesApi,
...
+                optionalParams.fold(BadRequest(views.html.register.email_verify_error("hts.email-verification.email-verify-error.bad-continue-url.content"))){ params ⇒
```
adwelly: We have a single page with a variety of error messages displayable. Plugging in the message key to denote the actual message is common in other repos I believe, and has the advantage of minimising the amount of code in the views where the tooling is less sophisticated.

app/uk/gov/hmrc/helptosavefrontend/views/register/update_email_address.scala.html
adwelly: There is not. Will change this.

Jackie1988: I have removed the Option tag

app/uk/gov/hmrc/helptosavefrontend/util/FormHelpers.scala
adwelly: It's a utility function written in scala whereas all the files in the view hierarchy are in Twirl. It could be moved of course - is this what we want?

app/uk/gov/hmrc/helptosavefrontend/util/FormHelpers.scala
adwelly: Yes - we should use this.

app/uk/gov/hmrc/helptosavefrontend/controllers/UpdateEmailAddressController.scala
adwelly: I think we should try to minimise the amount of code in views because the tooling is so primitive - certainly passing in message keys is a standard pattern in other projects.

app/uk/gov/hmrc/helptosavefrontend/connectors/EmailVerificationConnector.scala
Jackie1988: ok

app/uk/gov/hmrc/helptosavefrontend/util/EmailVerificationParams.scala
Jackie1988: Ok I have replaced : with §

app/uk/gov/hmrc/helptosavefrontend/controllers/UpdateEmailAddressController.scala
Jackie1988: Ok updated

AndyHWChung:
LGTM

PR #172, opened 2017-08-24
Stephen test branch

Stephen-Kam:
Modified the acceptance tests, notable changes:
-Enabled functionality to run in different environments by passing in an environment property
-Added/corrected tests to run on latest version of the frontend
-Refactored a lot of code, e.g. removed all references in each function that required an implicit webdriver and just had one implicit webdriver
-Enabled tests to run via sbt using the selenium:test hook


selenium-system-test/src/test/scala/hts/steps/Steps.scala
```diff
@@ -24,16 +24,17 @@ import hts.driver.Driver
...
 import org.scalatest.Matchers
 
+
 trait Steps extends ScalaDsl with EN with Matchers {
 
   import Steps._
 
   /** Tries to get the value of [[_driver]] - will throw an exception if it doesn't exist */
-  implicit def getDriverUnsafe: WebDriver = _driver.getOrElse(sys.error("Driver does not exist"))
+  implicit def driver: WebDriver = _driver.getOrElse(sys.error("Driver does not exist"))
```
AndyHWChung: this shouldn't be used outside of this trait, it's unsafe in that it will throw if `_driver` doesn't exist - I only put it in here because things extending `Steps` had the guarantee that  a test was being run so the `Before` and `After` blocks were being called below

project/FrontendBuild.scala
```diff
@@ -29,7 +29,7 @@ object FrontendBuild extends Build with MicroService {
   def test(scope: String = "test") = Seq(
     "uk.gov.hmrc" %% "hmrctest" % "2.3.0" % scope,
     "org.scalatest" %% "scalatest" % "3.0.1" % scope,
-    "org.pegdown" % "pegdown" % "1.6.0" % scope,
+    "org.pegdown" % "pegdown" % "1.4.2" % scope,
```
AndyHWChung: is there a particular reason we're downgrading? Are we even using this library?

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -17,26 +17,46 @@
...
 
 import hts.utils.Configuration
-import org.openqa.selenium.{By, WebDriver}
+import org.openqa.selenium.By
 
-object AuthorityWizardPage {
+object AuthorityWizardPage extends WebPage {
 
-  def goToPage()(implicit driver: WebDriver): Unit =
+  var tempNino: String = ""
```
AndyHWChung: do we really need to store the nino here - the only place it's being set is `authenticateUser` where the nino is already known. Is there a reason why the code calling this object will have lost that nino in between `authenticateUser` and `getNino`?

AndyHWChung:
LGTM

PR #182, opened 2017-08-31
Introduced wart remover

AndyHWChung:


AndyHWChung:
warts are here: http://www.wartremover.org/doc/warts.html. I disabled these:

```scala 
     Wart.DefaultArguments,
      Wart.FinalCaseClass,
      Wart.FinalVal,
      Wart.ImplicitConversion,
      Wart.ImplicitParameter,
      Wart.LeakingSealed,
      Wart.Nothing,
      Wart.Overloading,
      Wart.Var
```
and in tests I've further disabled these:

```scala
Wart.Any, Wart.Equals, Wart.Null, Wart.NonUnitStatements, Wart.PublicInference
```

because they were incompatible with various libraries we use in test

AndyHWChung:
I disabled `Nothing` because if we had this in the code:
```scala
def f(i: Int): Either[String,Int] = if(i == 0) Left("Uh oh") else Right(i)
```
it would infer the right type on the `Left` and the left type on the `Right` as `Nothing` so we would have to write something like this everytime we use `Left`s and `Right`s:
```scala
def f(i: Int): Either[String,Int] = if(i == 0) Left[String,Int]]("Uh oh") else Right[String,Int](i)
```
but this is annoying

app/uk/gov/hmrc/helptosavefrontend/util/Toggles.scala
```diff
@@ -16,58 +16,50 @@
...
 package uk.gov.hmrc.helptosavefrontend.util
 
-import java.time.Instant
-
+import com.typesafe.config.Config
 import play.api.{Configuration, Logger}
 
 object Toggles {
 
-  case class FEATURE[A](name: String, conf: Configuration, unconfiguredVal: Option[A], logger: Logger) {
```
AndyHWChung: this file was activating the `Throw` wart and when it was being used it was activating the `NonUnitStatements` wart. Since these two warts were inherent in the design I've replaced it with a cleaner implementation with no warts

app/uk/gov/hmrc/helptosavefrontend/config/FrontendWiring.scala
```diff
@@ -16,30 +16,48 @@
...
+class WSHttpExtension extends WSHttp {
+
+  override val hooks: Seq[HttpHook] = NoneRequired
 
   /**
    * Returns a [[Future[HttpResponse]] without throwing exceptions if the status us not `2xx`. Needed
    * to replace [[GET]] method provided by the hmrc library which will throw exceptions in such cases.
    */
   def get(url: String)(implicit rhc: HeaderCarrier): Future[HttpResponse] = withTracing(GET_VERB, url) {
+    // Future.successful(1)
```
sureshhmrc: can we remove this comment ?

app/uk/gov/hmrc/helptosavefrontend/config/FrontendWiring.scala
```diff
@@ -60,22 +78,17 @@ trait WSHttpExtension extends WSGet with WSPost {
 }
 
 @Singleton
-class WSHttp extends WSGet with WSPut with WSPost with WSDelete with AppName with RunMode with WSHttpExtension {
```
sureshhmrc: where are the `AppName` and  `RunMode` compensated ?  or we don't need them ?

app/uk/gov/hmrc/helptosavefrontend/controllers/EligibilityCheckController.scala
```diff
@@ -137,10 +138,11 @@ class EligibilityCheckController @Inject() (val messagesApi:             Message
...
 
           // set the ITMP flag here but don't worry about the result
-          helpToSaveService.setITMPFlag(nino).fold(
-            e ⇒ logger.warn(s"Could not set ITMP flag for user $nino: $e"),
-            _ ⇒ logger.info(s"Set ITMP flag for user $nino")
-          )
+          helpToSaveService.setITMPFlag(nino).value.onComplete{
+            case Failure(e)        ⇒ logger.warn(s"Could not set ITMP flag, future failed: ${e.getMessage}")
+            case Success(Left(e))  ⇒ logger.warn(s"Could not set ITMP flag: $e")
+            case Success(Right(_)) ⇒ logger.info(s"Set ITMP flag for user $nino")
```
sureshhmrc: the mesage could be `Successfully set ITMP flag for user $nino`

app/uk/gov/hmrc/helptosavefrontend/util/Toggles.scala
```diff
@@ -16,58 +16,50 @@
...
-          logger.info("FEATURE: " + name + (if (b) " enabled" else " disabled"))
-          FEATURE_THEN(name, b, unconfiguredVal, logger)
-        case None ⇒ throw new RuntimeException(s"FEATURE($name) is not present in configuration file - misconfigured")
-      }
-    }
-  }
-
-  object FEATURE {
-    def apply[A](name: String, conf: Configuration): FEATURE[A] = FEATURE[A](name: String, conf: Configuration, None, Logger(name))
+  /**
```
sureshhmrc: i prefer to get rid of this Toggles Feature completely :)

app/uk/gov/hmrc/helptosavefrontend/views/introduction/navigation.scala.html
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  *@
 
+@import cats.instances.string._
+@import cats.syntax.eq._
+
```
sureshhmrc: can we disable warts for all views so that views cane be made free from cats ? :) 
not a big deal though..

selenium-system-test/src/test/scala/hts/driver/Driver.scala
```diff
@@ -18,6 +18,8 @@ package hts.driver
 
 import java.util.concurrent.TimeUnit
 
+import cats.instances.string._
```
sureshhmrc: Also I don't care warts rules for selenium code :)

test/uk/gov/hmrc/helptosavefrontend/controllers/IntroductionControllerSpec.scala
```diff
@@ -31,7 +31,7 @@ import scala.concurrent.Future
 
 class IntroductionControllerSpec extends AuthSupport {
 
-  implicit val timeout = Timeout(5, SECONDS)
+  implicit val timeout: Timeout = Timeout(5, SECONDS)
```
sureshhmrc: can we move this timeout to base class TestSupport or AuthSupport instead of declaring in each spec ?

sureshhmrc:
@AndyHWChung  overall looks ok , added minor comments... feel free to merge once you add any changes...

app/uk/gov/hmrc/helptosavefrontend/config/FrontendWiring.scala
AndyHWChung: whoops, yeah I was using that to test the warts library

app/uk/gov/hmrc/helptosavefrontend/config/FrontendWiring.scala
AndyHWChung: looks like we never needed them - at least everything still compiles when they're not there

app/uk/gov/hmrc/helptosavefrontend/views/introduction/navigation.scala.html
AndyHWChung: since there's scala code in the views I think it's probably a good idea to keep it there so that the code quality is consistent across the whole code base - similarly for the selenium tests as well

app/uk/gov/hmrc/helptosavefrontend/controllers/EnrolmentCheckBehaviour.scala
```diff
@@ -48,11 +49,13 @@ trait EnrolmentCheckBehaviour {
         // if the user is enrolled but the itmp flag is not set then just
         // start the process to set the itmp flag here without worrying about the result
         if (!itmpHtSFlag) {
-          helpToSaveService.setITMPFlag(nino).fold(
-            e ⇒ logger.warn(s"Could not start process to set ITMP flag for user $nino: $e"),
-            _ ⇒ logger.info(s"Process started to set ITMP flag for user $nino")
-          )
+          helpToSaveService.setITMPFlag(nino).value.onComplete{
+            case Failure(e)        ⇒ logger.warn(s"Could not start process to set ITMP flag for user $nino, future failed: $e")
```
tim-holbrook: For all NINO log recordings, pls can you start the msg "For NINO, xxxx" as this will ensure all our log msgs are consistent in this respect, to help any post-live support

AndyHWChung: sure - but aren't we going to remove all NINO logging in the frontend at some point? Or are we not doing that anymore?

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -53,19 +53,20 @@ object NSIConnector {
...
   override def createAccount(userInfo: NSIUserInfo)(implicit hc: HeaderCarrier, ex: ExecutionContext): Future[SubmissionResult] = {
     import uk.gov.hmrc.helptosavefrontend.util.Toggles._
 
     logger.info(s"Trying to create an account for ${userInfo.nino} using NSI endpoint $nsiUrl")
 
-    FEATURE("log-account-creation-json", conf) enabled () thenDo {
-      logger.info(s"CreateAccount json for ${userInfo.nino} is ${Json.toJson(userInfo)}")
-    }
+    FEATURE("log-account-creation-json", conf, logger).thenOrElse(
+      logger.info(s"CreateAccount json for ${userInfo.nino} is ${Json.toJson(userInfo)}"),
```
tim-holbrook: Pls can you change to "For NINO ${userInfo.nino} xxxx"

app/uk/gov/hmrc/helptosavefrontend/controllers/EligibilityCheckController.scala
```diff
@@ -137,10 +138,11 @@ class EligibilityCheckController @Inject() (val messagesApi:             Message
...
-            e ⇒ logger.warn(s"Could not set ITMP flag for user $nino: $e"),
-            _ ⇒ logger.info(s"Set ITMP flag for user $nino")
-          )
+          helpToSaveService.setITMPFlag(nino).value.onComplete{
+            case Failure(e)        ⇒ logger.warn(s"For NINO [$nino]: Could not set ITMP flag, future failed: ${e.getMessage}")
+            case Success(Left(e))  ⇒ logger.warn(s"For NINO [$nino]: Could not set ITMP flag: $e")
+            case Success(Right(_)) ⇒ logger.info(s"For NINO [$nino]: Successfully set ITMP flag for user")
+          }
 
           Ok("You've already got an account - yay!!!")
```
tim-holbrook: Do we need to chat about what the final msg will be here?

AndyHWChung: this is a stub page - this will actually be a redirect to NS&I but we don't have that yet so this is a placeholder page

PR #183, opened 2017-08-31
Hts 310

Stephen-Kam:
The shell script to be able to run via tags wasn't working correctly, this has now been corrected
Modified the auth wizard object due to the new ITMP stub changes


selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -52,4 +63,37 @@ object AuthorityWizardPage extends WebPage {
   def submit()(implicit driver: WebDriver): Unit =
     find(cssSelector("input.button")).get.underlying.click()
 
+  def setGivenName(givenName: String)(implicit driver: WebDriver): Unit =
+    find(name("itmp.givenName")).get.underlying.sendKeys(givenName)
```
AndyHWChung: avoid `.get` on `Option` as this will throw when the `Option` is not defined. For this case you could do something like this instead:
```scala
find(name("itmp.givenName")).foreach(_.underlying.sendKeys(givenName))
```

AndyHWChung: the same thing applies to the other methods below

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -31,15 +31,15 @@ object AuthorityWizardPage extends WebPage {
     setConfidenceLevel(confidence)
     setCredentialStrength(credentialStrength)
     setNino(nino)
-    setGivenName("a")
-    setMiddleName("a")
-    setFamilyName("a")
+    setGivenName("GivenName")
+    setMiddleName("MiddleName")
```
AndyHWChung: (we don't actually use the middle name btw so this is probably not needed)

PR #186, opened 2017-09-01
Hts 310

Stephen-Kam:
modified shell script to take the driver location as optional and also to work as intended

-If you don't add in a driver location then it will use the locally stored one based on your operating system

-If running in Jenkins, then a driver location can be added to ensure it picks up the correct driver location (/usr/local/bin/chromedriver)


selenium-system-test/src/test/scala/hts/driver/Driver.scala
```diff
@@ -62,8 +62,8 @@ object Driver {
   private val driverDirectory: String = "selenium-system-test/drivers"
 
   private def createChromeDriver(): WebDriver = {
-    if (Option(systemProperties.getProperty("webdriver.chrome.driver")).isEmpty) {
-      val _ = if (isMac) {
+    if (systemProperties.getProperty("webdriver.chrome.driver") == null) {
```
AndyHWChung: We should avoid using `null` wherever possible - the wrapping of the return value in the `Option` and checking if the result is `None` is the proper way to work with `null` values in scala since
```scala
Option(null) = None
```
Avoiding `null` now will get you in the practice of not using it in the future. See this info here: https://alvinalexander.com/scala/scala-null-values-option-uninitialized-variables

run_selenium_system_test.sh
```diff
@@ -40,9 +40,25 @@ function get_tags {
 
 # Return the necessary java options as a single string. Each java option is surrounded by double quotes and
 # separated by a comma
+
+function get_driver_location {
+local driverLocation=$@
+
+driver+=$driverLocation
+
+echo $driver
```
AndyHWChung: is it more useful to echo a more descriptive string?

selenium-system-test/src/test/scala/hts/driver/Driver.scala
AndyHWChung: This would have been picked up by WartRemover but I had to disable the null checks because scalamock was giving false positives

AndyHWChung:
LGTM

PR #191, opened 2017-09-08
Removed embedded browser driver files from project

MarcusQA:


selenium-system-test/src/test/scala/hts/steps/EmailSteps.scala
```diff
@@ -16,33 +16,25 @@
...
-  When("""^they wish to change their email$"""){ () ⇒
-
-  }
-
-  When("""^they enter in a new email and confirm$"""){ () ⇒
-
-  }
-
-  When("""^they indicate that they wish to provide an alternative email address$"""){ () ⇒
+  var nino: Option[String] = None
```
AndyHWChung: do we want to have a `nino` var here given that we can use the current NINO that's been generated in the `NINOGenerator`? This may not be a fix to make in this PR though since Jackie's working on that stuff now

AndyHWChung:
LGTM

PR #195, opened 2017-09-10
Re-added test for confirming applicant details and refactored existin…

MarcusQA:
…g tests


selenium-system-test/src/test/resources/features/confirmDetails/Applicant confirms details.feature
```diff
@@ -0,0 +1,14 @@
...
+    Given an applicant has the following details:
+      |field          |value            |
+      |first name     |FirstName        |
+      |last name      |LastName         |
+      |NINO           |<auto-generated> |
+      |date of birth  |20/12/1980       |
+      |email address  |user@test.com    |
+
+    When an applicant passes the eligibility check
+    Then they see their details
```
AndyHWChung: instead of using the `Given` step to set `vars` in the `UserDetailsStep` class is it not possible to do something like:
```
Scenario: Applicant confirms details
When an applicant passes the eligibility check with the following details:
       |field              |value                    |
       |first name     |FirstName            |
       |last name      |LastName            |
       |NINO           |<auto-generated> |
       |date of birth  |20/12/1980         |
       |email address  |user@test.com    |
  Then they see the following details:
       |field              |value                    |
       |first name     |FirstName            |
       |last name      |LastName            |
       |NINO           |<auto-generated> |
       |date of birth  |20/12/1980         |
       |email address  |user@test.com    |
```

This avoids `vars` floating around which means if we want to re-use these step definitions its clearer what parameters are going to be used

AndyHWChung:
LGTM

PR #196, opened 2017-09-11
added more timer metrics

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -70,44 +68,55 @@ class NSIConnectorImpl @Inject() (conf: Configuration, auditor: HTSAuditor, metr
...
+        logger.warn(s"Received 500 from NSI, failed to create account for $nino as there was an " +
+          s"internal server error ${timeString(time)}")
+        handleBadRequestResponse(response)
+
+      case Status.SERVICE_UNAVAILABLE ⇒
+        logger.warn(s"Received 503 from NSI, failed to create account for $nino as NSI " +
+          s"service is unavailable ${timeString(time)}")
+        handleBadRequestResponse(response)
+
+      case other ⇒
```
sureshhmrc: this is still an error, don't we want to increment counters here ?

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -70,44 +68,55 @@ class NSIConnectorImpl @Inject() (conf: Configuration, auditor: HTSAuditor, metr
       ()
     )
 
-    val timeContext: Timer.Context = timer.time()
+    val timeContext: Timer.Context = metrics.nsiAccountCreationTimer.time()
```
sureshhmrc: in the unit tests, now we should be able to verify  whether the metrics are working correctly since metrics is injected

app/uk/gov/hmrc/helptosavefrontend/connectors/SessionCacheConnector.scala
```diff
@@ -21,6 +21,7 @@ import com.google.inject.{ImplementedBy, Inject, Singleton}
 import play.api.libs.json.{Reads, Writes}
```
sureshhmrc: are we not doing metrics for other connectors  ?

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
AndyHWChung: I was a bit lazy with doing unit tests for metrics since they aren't part of the production logic per se, but I can go ahead and put in unit tests if we really want them there

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
```diff
@@ -70,44 +68,55 @@ class NSIConnectorImpl @Inject() (conf: Configuration, auditor: HTSAuditor, metr
...
 
     httpProxy.post(nsiUrl, userInfo, Map(nsiAuthHeaderKey → nsiBasicAuth))
       .map[SubmissionResult] { response ⇒
         val time = timeContext.stop()
 
         response.status match {
           case Status.CREATED ⇒
-            auditor.sendEvent(new ApplicationSubmittedEvent(appName, userInfo))
+            auditor.sendEvent(ApplicationSubmittedEvent(appName, userInfo))
```
sureshhmrc: can we add counters for success cases as well ?

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
AndyHWChung: I moved the increment of the counter above of the pattern match so it will get incremented in this case

app/uk/gov/hmrc/helptosavefrontend/connectors/SessionCacheConnector.scala
AndyHWChung: there's no reason not to - I'll stick them in there as well then

app/uk/gov/hmrc/helptosavefrontend/connectors/NSIConnector.scala
AndyHWChung: we can derive the success cases by doing the count of times timed minus the error count so we don't need an explicit success counter

PR #204, opened 2017-09-13
Implemented first endpoint to update emails from NS&I

AndyHWChung:


AndyHWChung:
https://jira.tools.tax.service.gov.uk/browse/HTS-432

PR #208, opened 2017-09-15
emails are now updated in mongo for account holders

AndyHWChung:


AndyHWChung:
https://jira.tools.tax.service.gov.uk/browse/HTS-462

PR #214, opened 2017-09-22
Hts 400 qa

MarcusQA:


selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
```diff
@@ -0,0 +1,117 @@
...
+      | address line 1 | 17 Ålfotbreen            |
+      | address line 2 | Grünerløkka              |
+      | address line 3 | Bodø                     |
+      | address line 4 | Hørdy-Gürdy4             |
+      | address line 5 | Hørdy-Gürdy5             |
+      | postcode       | 19023                    |
+      | country code   | 02                       |
+
+    When they choose to go ahead with creating an account
+    Then they see that the account is created
```
sureshhmrc: i see only happy path scenarios above, cant we test non-happy path cases ?

selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
```diff
@@ -0,0 +1,117 @@
...
+
+
+  Scenario: Air-gap 11 - Populate forename, surname, address lines (phone number) and email with accented characters
+    Given an applicant has the following details:
+      | field          | value                    |
+      | first name     | AirGapRené Chloë         |
+      | last name      | ElevenO'Connor-Jørgensen |
+      | NINO           | <eligible>               |
```
sureshhmrc: why only `<eligible>` in all scenarios ?

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -16,13 +16,19 @@
...
 
-import hts.utils.Configuration
+import java.time.format.DateTimeFormatter
+
+import hts.utils.{Configuration, TestUserInfo}
 import org.openqa.selenium.WebDriver
+import uk.gov.hmrc.helptosavefrontend.models.{Address, UserInfo}
+
+import scala.annotation.tailrec
+import scala.util.Try
```
sureshhmrc: unused `Try` ?

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -44,7 +50,24 @@ object AuthorityWizardPage extends WebPage {
...
 
-  def goToPage()(implicit driver: WebDriver): Unit =
+  def enterUserDetails(confidence: Int, credentialStrength: String, userInfo: TestUserInfo)(implicit driver: WebDriver): Unit = {
+    navigate()
+    setConfidenceLevel(confidence)
+    setCredentialStrength(credentialStrength)
+
+    setAddressLines(userInfo.address)
+
+    userInfo.address.postcode.foreach(setPostCode)
```
sureshhmrc: can we use map instead of foreach on an Option ?

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -95,4 +118,26 @@ object AuthorityWizardPage extends WebPage {
   def setCountryCode(countryCode: String)(implicit driver: WebDriver): Unit =
     find(name("itmp.address.countryCode")).foreach(_.underlying.sendKeys(countryCode))
 
+  private def setAddressLines(address: Address)(implicit driver: WebDriver): Unit = {
```
sureshhmrc: can we make this logic simple ?

selenium-system-test/src/test/scala/hts/steps/ConfirmDetailsSteps.scala
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package hts.steps
+
+import java.time.format.DateTimeFormatter
+
+import cucumber.api.DataTable
+import hts.pages.{AuthorityWizardPage, EligiblePage}
+import hts.utils.{Helpers, ScenarioContext, TestUserInfo}
+import uk.gov.hmrc.helptosavefrontend.models.UserInfo
```
sureshhmrc: unused import ?

selenium-system-test/src/test/scala/hts/steps/CreateAccountSteps.scala
```diff
@@ -18,29 +18,31 @@ package hts.steps
 
 import hts.pages._
 import hts.pages.registrationPages._
-import hts.utils.{Configuration, NINOGenerator}
+import hts.utils.{Configuration, ScenarioContext}
+import uk.gov.hmrc.helptosavefrontend.models.UserInfo
```
sureshhmrc: unused import ?

selenium-system-test/src/test/scala/hts/utils/Helpers.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package hts.utils
+
+import hts.pages.Page
+import org.openqa.selenium.WebDriver
+
+object Helpers {
+  def isTextOnPage(text: String)(implicit driver: WebDriver): Boolean = {
+    text.r.findAllIn(Page.getPageContent).length >= 1
+  }
```
sureshhmrc: you can simply say 
```  def isTextOnPage(text: String)(implicit driver: WebDriver): Boolean = text.r.findAllIn(Page.getPageContent).nonEmpty```

File: selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala (hunk 4 above)
Comment (AndyHWChung) [reply selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala]: yeah - `map` will basically do the same thing as `foreach` if the function you're using in `map` returns `Unit`. `foreach` is used for side-effecting functions which return `Unit` whereas `map is used to transform - this is why `foreach` returns `Unit` whereas `map` returns another `Option`. Here we're dealing with side-effecting functions which is why `foreach` is probably more appropriate here

File: selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala (hunk 5 above)
Comment (AndyHWChung) [reply selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala]: we had something like 
```scala
Try(address.lines(0)).toOption.foreach(setAddressLine1)
Try(address.lines(1)).toOption.foreach(setAddressLine2)
Try(address.lines(2)).toOption.foreach(setAddressLine3)
...
```
which uses the unsafe accessor methods on the `List` which were generating compile-time warnings. The code here was a way to avoid those warninings

selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
MarcusQA: I wanted to highlight that the user is eligible for HtS. At present, only "eligible" exists as a tag in the feature files to describe a NINO belonging to an eligible user - but I plan to introduce other tags.

selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
MarcusQA: We can't test the unhappy paths yet because we don't have the ability to make mandatory fields blank - this issue is blocking https://jira.tools.tax.service.gov.uk/browse/HTS-461 too.

selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
Jackie1988: I'm going to merge this in, we can add more scenarios later

selenium-system-test/src/test/resources/features/airGapTests/Air-gap tests for testing with NS&I.feature
Jackie1988: I'm going to merge this in, we can add unhappy cases later

PR #218, opened 2017-09-26
Hts 222

Jackie1988:


run_selenium_system_test.sh
```diff
@@ -9,7 +9,7 @@ tags    - Space separated list of tags. Only run tests with the given
           tags. Not specifying any tags will run all tests.
```
MarcusQA: On line 6 of  run_selenium_system_test.sh can we include [zap-chrome], please?

project/MicroService.scala
```diff
@@ -22,6 +22,8 @@ trait MicroService {
   lazy val plugins: Seq[Plugins] = Seq.empty
   lazy val playSettings: Seq[Setting[_]] = Seq.empty
 
+  val hmrcRepoHost = java.lang.System.getProperty("hmrc.repo.host", "https://nexus-dev.tax.service.gov.uk")
```
MarcusQA: Is this used if you're running the zap test locally?

Jackie1988: I don't think this is used at all atm, I will check and then remove.

run_selenium_system_test.sh
Jackie1988: Yeah sure

PR #221, opened 2017-09-26
selenium test changes

shaunkenclark:
Changes to failing selenium tests that were not executing as required.


selenium-system-test/src/test/scala/hts/pages/WebPage.scala
```diff
@@ -34,7 +34,24 @@ trait WebPage extends Matchers
...
+  def nextPage()(implicit driver: WebDriver): Unit = {
+    driver.findElement(By.id("get-help")).sendKeys(Keys.chord(Keys.CONTROL, Keys.END))
+    find(CssSelectorQuery(".page-nav__link.page-nav__link--next")).foreach(click.on)
+  }
+
+  /*
+   * CONFIRM CURRENT PAGE INFO
+   */
+
+  def expectedUrl: Option[String]
```
AndyHWChung: Why is the return type `Option[String]`? Does `None` mean checks against this expected URL should always return `true`? The way you have overriden it, this seems to always return either `Some` with a value in it or `Some` with an empty `String` in it. Is the `Some("")` actually your `None` case?

shaunkenclark: Hi Andy,

I set it as Option[String] so that I could set them as None in the event that the page or element to which the comparison was taking place wouldn't fail if that particular object didn't have those elements (admittedly that's probably not necessary for URL as that will always be present). I'm still pretty new to Scala however, so I'm open to any suggestions or corrections. 

The ones currently just set to Some("") are pages which never use the pageInfoIsCorrect method. In future I am of the position that these details should always be checked when a new page is loaded.

shaunkenclark:
Changed `expectedUrl`, `expectedPageTitle`, and `expectedPageHeader` to direct `String` types instead. 

`currentUrl` is a scalatest method that should always return a value (at least in respect to our tests).
`pageTitle` is also a scalatest method and returns an empty `String`, `""` in the event no title is found. 
I created an additional method `pageHeading` to do the same for the heading, return an empty `String` in the event none is found. 

I also instantiated the values in the `Page.scala` to class so that they are not required to be overridden in child classes if not needed.

selenium-system-test/src/test/scala/hts/steps/EligibilitySteps.scala
```diff
@@ -31,7 +31,8 @@ class EligibilitySteps extends Steps {
   }
 
   Then("""^they see that they are eligible for Help to Save$""") { () ⇒
-    on(EligiblePage)
+    //on(EligiblePage)
```
AndyHWChung: do we need this comment?

selenium-system-test/src/test/scala/hts/steps/EligibilitySteps.scala
```diff
@@ -43,6 +44,8 @@ class EligibilitySteps extends Steps {
   }
 
   Then("""^they see that they are NOT eligible for Help to Save$""") { () ⇒
-    on(NotEligiblePage)
+    //on(NotEligiblePage)
```
AndyHWChung: and do we need this one?

selenium-system-test/src/test/scala/hts/steps/Steps.scala
```diff
@@ -34,6 +34,9 @@ private[steps] trait Steps extends ScalaDsl with EN with Matchers {
   /** Tries to get the value of [[_driver]] - will throw an exception if it doesn't exist */
   implicit def driver: WebDriver = _driver.getOrElse(sys.error("Driver does not exist"))
 
+  /*
+   * method "on" not currently in use
```
AndyHWChung: if it's not being used do we need to get rid of this method?

shaunkenclark: We shouldn't need any of those commented out or this method, I just neglected to remove them. I will do so now.

AndyHWChung:
LGTM

PR #225, opened 2017-09-28
Added headless chrome as a browser

MarcusQA:


selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -72,8 +72,10 @@ object AuthorityWizardPage extends WebPage {
   def setRedirect(url: String)(implicit driver: WebDriver): Unit =
     find(name("redirectionUrl")).foreach(_.underlying.sendKeys(url))
 
-  def setNino(nino: String)(implicit driver: WebDriver): Unit =
+  def setNino(nino: String)(implicit driver: WebDriver): Unit = {
+    println("NINO entered in auth wizard: " + nino)
```
sureshhmrc: are we using printlns ? if this is just for logging, thats fine

selenium-system-test/src/test/scala/hts/utils/ScenarioContext.scala
```diff
@@ -98,6 +98,8 @@ private[utils] trait NINOGenerator {
     ineligibleNino
   }
 
-  def currentEligibleNINO(): String = toEligible(current)
-
+  def currentNINO(): String = {
+    println("Current NINO: " + current)
```
sureshhmrc: another println()

PR #229, opened 2017-10-02
Made NINO logging consistent

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -68,24 +69,24 @@ class RegisterController @Inject() (val messagesApi:           MessagesApi,
 
           result.fold[Result](
             { e ⇒
-              logger.warn(s"For NINO [$nino]: Could not write confirmed email: $e")
+              logger.warn(s"For NINO [$nino]: Could not write confirmed email: $e", nino)
```
Jackie1988: This doesn't seem to follow your new logging 'pattern'

AndyHWChung: Yup, I missed that one - will change it now!

PR #246, opened 2017-10-12
HTS-529: Pass correct Authorization header

sureshhmrc:


sureshhmrc:
https://jira.tools.tax.service.gov.uk/browse/HTS-529

PR #248, opened 2017-10-13
HTS-519:  Fix the missing continueURL param in the IV result

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/controllers/IvController.scala
```diff
@@ -45,6 +45,7 @@ class IvController @Inject() (val sessionCacheConnector: SessionCacheConnector,
     //Will be populated if we arrived here because of an IV success/failure
     val journeyId = request.getQueryString("token").orElse(request.getQueryString("journeyId"))
     val allowContinue = true
+    val newIvUrl = ivUrl(continueURL)
```
MarcusQA: Just a question - does "new" in the variable name convey what the variable is doing?

sureshhmrc: not like that - i meant to create . a new iv url val here...

sureshhmrc: i used the name as we already have a method name `ivUrl(....)` so it not allow me to write 
`val ivUrl = ivUrl(continueURL)` (edited)

AndyHWChung:
LGTM

PR #258, opened 2017-10-18
HTS-502: Auditing fixes 

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/connectors/HelpToSaveConnector.scala
```diff
@@ -123,29 +123,15 @@ class HelpToSaveConnectorImpl @Inject() (http: WSHttp)(implicit ec: ExecutionCon
...
-      case other ⇒
-        Left(s"Could not parse eligibility result code '$other'. Result string was '$reason'")
-
+  // scalastyle:off magic.number
+  private def validate(response: EligibilityCheckResponse): Either[String, EligibilityCheckResponse] = {
+    (response.resultCode, response.reasonCode) match {
+      case (1, 6) | (1, 7) | (1, 8)          ⇒ Right(response) // eligible
+      case (2, 2) | (2, 3) | (2, 4) | (2, 5) ⇒ Right(response) // non-eligible
+      case (3, 1)                            ⇒ Right(response) // account already opened
+      case invalid                           ⇒ Left(s"Could not parse eligibility response, unknown combination (resultCode, reasonCode) = ${(invalid._1, invalid._2)}")
```
tim-holbrook: I'm just wondering here if we need to check the 2nd number, as this increases validation coverage, but also increases fragility.  The whole idea of changing Account Already Created to 3,1 was to just use the resultCode for our three-way outcome.

test/uk/gov/hmrc/helptosavefrontend/models/HTSEventSpec.scala
```diff
@@ -70,41 +70,31 @@ class HTSEventSpec extends TestSupport {
...
-    }
-
-    "be created with the eligible tag set false, a reason tag, and the nino" in {
-      val reason = "reason"
-      val event = EligibilityResult(validNSIUserInfo.nino, reason, isEligible = false)(new HeaderCarrier)
-      event.value.detail.size shouldBe 3
-      event.value.detail.exists(x ⇒ x._1 === "nino" && x._2 === validNSIUserInfo.nino) shouldBe true
-      event.value.detail.exists(x ⇒ x._1 === "eligible" && x._2 === "false") shouldBe true
-      event.value.detail.exists(x ⇒ x._1 === "reason" && x._2 === reason) shouldBe true
+      event.value.detail.exists(x ⇒ x._1 === "result" && x._2 === "3, 1, meaning 'blah', 'already created'") shouldBe true
```
tim-holbrook: meaning 'HtS account already exists', 'HtS account was previously created'

app/uk/gov/hmrc/helptosavefrontend/connectors/HelpToSaveConnector.scala
```diff
@@ -37,7 +37,7 @@ import scala.util.control.NonFatal
 @ImplementedBy(classOf[HelpToSaveConnectorImpl])
 trait HelpToSaveConnector {
 
-  def getEligibility()(implicit hc: HeaderCarrier): EitherT[Future, String, EligibilityCheckResult]
+  def getEligibility()(implicit hc: HeaderCarrier): EitherT[Future, String, EligibilityCheckResponse]
```
AndyHWChung: The problem with returning `EligibilityCheckResponse` here is that the code in front is now too tightly coupled with  DES's interface with us. That interface is not very code friendly since it relies on an `Int` to convey eligibility, which is why later on in the code you have to do a match on `Ints` which is arbitrary when looked at in isolation. By converting the response we get from DES to something more sensibly typed it makes sure that we are less coupled to that interface and the code is more easily understood after this point.

app/uk/gov/hmrc/helptosavefrontend/models/EligibilityCheckResult.scala
```diff
@@ -1,93 +0,0 @@
...
- * limitations under the License.
- */
-
-package uk.gov.hmrc.helptosavefrontend.models
-
-sealed trait EligibilityCheckResult
-
-object EligibilityCheckResult {
-
-  case class Eligible(reason: EligibilityReason) extends EligibilityCheckResult
```
AndyHWChung: if we don't care about the individual reasons we could have:
```scala
  case class Eligible(description: String, code: Int) extends EligibilityCheckResult
  case class Ineligible(description: String, code: Int) extends EligibilityCheckResult
  case class AlreadyHasAccount(description: String, code: Int) extends EligibilityCheckResult
```

PR #286, opened 2017-10-31
implemented lock for health test

AndyHWChung:


project/FrontendBuild.scala
```diff
@@ -18,7 +18,8 @@ object FrontendBuild extends Build with MicroService {
     "com.github.kxbmap" %% "configs" % "0.4.4",
     "com.eclipsesource" %% "play-json-schema-validator" % "0.8.9",
     "uk.gov.hmrc" %% "play-whitelist-filter" % "2.0.0",
-    "uk.gov.hmrc" %% "zap-automation" % "0.15.0"
```
Jackie1988: Why has this been removed?

app/uk/gov/hmrc/helptosavefrontend/util/lock/Lock.scala
```diff
@@ -0,0 +1,132 @@
...
+import uk.gov.hmrc.lock.{ExclusiveTimePeriodLock, LockMongoRepository, LockRepository}
+import uk.gov.hmrc.helptosavefrontend.util.lock.LockProvider.ExclusiveTimePeriodLockProvider
+import uk.gov.hmrc.helptosavefrontend.util.toFuture
+
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
+
+/**
+ * This is an `Actor` which handles changing of state when a lock is acquired or a lock
```
tim-holbrook: Please can you add ... The objective of the Lock is to provide a single instance of a resource that is used in a multi-instance microservice environment.

....also, to mention what property name controls the period time with an example setting

conf/application.conf
```diff
@@ -349,6 +349,11 @@ health {
     poll-count-between-updates = 20
     poll-count-failures-to-alert = 3
     poll-count-between-pager-duty-alerts = 20
+    # in order to make sure that only one health check is running across the different instances
+    # of this application in each environment a lock needs to be acquired before the health check
+    # will start. This parameter controls how long this lock is valid for and when the health check
+    # will try to acquire the lock again
+    lock-period = 65 minutes
```
tim-holbrook: lock-reacquire-period ?

project/FrontendBuild.scala
AndyHWChung: it's already being brought in as a test dependency below - I didn't think we needed it outside of testing

app/uk/gov/hmrc/helptosavefrontend/util/lock/Lock.scala
AndyHWChung: That comment doesn't belong in the scaladoc for `Lock` since that actor uses the `LockProvider` trait which is generic and does not know anything about mongo lock. I've added documentation in `ExclusiveTimePeriodLockProvider` though to make it clearer what the intention is

AndyHWChung: For the `Lock` actor the lock duration is defined by the `LockProvider` being passed in - I've made this clear in the documentation now. The `Lock` actor does not know anything about how the lock duration is defined in the `LockProvider` since `LockProvider` is generic

tim-holbrook:
LGTM

PR #306, opened 2017-11-07
increased test coverage

AndyHWChung:


AndyHWChung:
Testing results in this log by scoverage now:
```
[info] Coverage is above minimum [96.79% > 92.0%]
```

sureshhmrc:
if we have `96.79%` now, then why not upping the min from `92.0%` to `95.0%` ? :)

PR #319, opened 2017-11-14
HTS-577: Implemented Deskpro Feedback functionality into our views

Jackie1988:


conf/messages
```diff
@@ -181,7 +181,7 @@ hts.helpers.sidebar.government-gateway=Government Gateway
 hts.helpers.sidebar.universal-credit=Universal Credit
 hts.helpers.sidebar.apply-for-universal-credit=Apply for Universal Credit
 hts.helpers.sidebar.working-tax-credit=Working Tax Credit
-hts.helpers.phaseBanner.feedback=This is a new service – your feedback will help us to improve it.
+hts.helpers.phaseBanner.feedback=This is a new service – your <a href="{0}">feedback</a> will help us to improve it.
```
AndyHWChung: Do we want HTML in our messages file? It hides it from the view file where the reset of the HTML is

Jackie1988: We don't need that anyway, so I have removed that along with the phaseBanner helper.

app/uk/gov/hmrc/helptosavefrontend/views/govuk_wrapper.scala.html
```diff
@@ -103,17 +102,23 @@
...
 
-
 @getHelpForm = @{uiHelpers.reportAProblemLink(appConfig.reportAProblemPartialUrl, appConfig.reportAProblemNonJSUrl)}
 
+@betaBanner = {
+@uiLayouts.betaBanner(false, appConfig.betaFeedbackUrlNoAuth, appConfig.betaFeedbackUrlNoAuth, true, true)
+}
+
+@serviceInfoContent = {
+@uiLayouts.serviceInfo(betaBanner, false, Option.empty, true )
```
AndyHWChung: is `Option.empty` not just `None`?

AndyHWChung:
LGTM

PR #320, opened 2017-11-15
HTS-580: Removed get help link from govuk_wrapper

Jackie1988:


app/uk/gov/hmrc/helptosavefrontend/views/govuk_wrapper.scala.html
```diff
@@ -49,13 +49,6 @@
 
 @headerNavLinks = {
         <li>
-            <a href="/get-help" id="get-help"
```
AndyHWChung: will we ever need it in the future? Is it safer to comment it out for now?

AndyHWChung: apparently we won't need this in the future so deleting is ok.

AndyHWChung:
LGTM

app/uk/gov/hmrc/helptosavefrontend/views/govuk_wrapper.scala.html
Jackie1988: Done

Jackie1988: oh ok

PR #348, opened 2017-11-29
Hts 598

Jackie1988:


selenium-system-test/src/test/scala/hts/steps/CreateAccountSteps.scala
```diff
@@ -64,8 +71,8 @@ class CreateAccountSteps extends Steps with Page {
     ApplyPage.clickSignInLink()
   }
 
-  When("""^they choose to create an account$""") { () ⇒
-    ConfirmDetailsPage.continue()
+  When("""^they click on accept and create an account$""") { () ⇒
+    Thread.sleep(4000)
```
AndyHWChung: do we really want a `Thread.sleep` here? There's almost always a better solution than `Thread.sleep`. The time you sleep for here may be enough for your local environment but it may require a different amount of time in others. The best solution is to find some sort of signal which signifies when the event you want to have happened has happened

selenium-system-test/src/test/scala/hts/steps/CreateAccountSteps.scala
```diff
@@ -88,30 +98,39 @@ class CreateAccountSteps extends Steps with Page {
...
     AuthorityWizardPage.enterUserDetails(200, "Strong", ScenarioContext.userInfo().getOrElse(sys.error))
     AuthorityWizardPage.setRedirect(EligiblePage.url)
     AuthorityWizardPage.submit()
     EligiblePage.startCreatingAccount()
-    ConfirmDetailsPage.continue()
+    SelectEmailPage.selectGGEmail()
+    SelectEmailPage.clickContinue()
     CreateAccountPage.createAccount()
+
+    //webDriverWait.Until(driver1 => ((IJavaScriptExecutor)driver).ExecuteScript("return document.readyState").Equals("complete"))
```
AndyHWChung: this stuff is probably not needed anymore right?

PR #376, opened 2017-12-07
HTS-589 Content and format for cannot change email try later

thebailers:
Update layout for cannot change email pages.


AndyHWChung:
LGTM

PR #380, opened 2017-12-07
Hts 330 take2

Jackie1988:


test/uk/gov/hmrc/helptosavefrontend/controllers/EligibilityCheckControllerSpec.scala
```diff
@@ -20,12 +20,15 @@ import java.time.LocalDate
 
 import cats.data.EitherT
 import cats.instances.future._
+//import org.joda.time.LocalDate
```
MarcusQA: Can we safely remove the commented-out line?

PR #383, opened 2017-12-08
Hts 490

fordie:


thebailers:
Front end bits look good to me.

AndyHWChung:
Did you run the unit tests?

AndyHWChung:
I ran the unit tests and I fixed one that was broken

PR #384, opened 2017-12-10
Fix selenium

MarcusQA:


selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -95,8 +95,10 @@ object AuthorityWizardPage extends Page {
   def setRedirect(url: String)(implicit driver: WebDriver): Unit =
     find(name("redirectionUrl")).foreach(_.underlying.sendKeys(url))
 
-  def setNino(nino: String)(implicit driver: WebDriver): Unit =
+  def setNino(nino: String)(implicit driver: WebDriver): Unit = {
+    println("NINO entered on auth wizard: " + nino)
```
AndyHWChung: at this point the NINO hasn't actually been entered. It would be more accurate to do:
```scala
find(name("nino")).foreach{ x =>
 x.underlying.sendKeys(nino)
println("...")
}
```
(`x` is not a good name above, I just can't remember what the type of it is right now)

selenium-system-test/src/test/scala/hts/steps/ConfirmDetailsSteps.scala
```diff
@@ -48,7 +48,8 @@ class ConfirmDetailsSteps extends Steps {
     val fullName = forename + " " + surname
 
     Helpers.isTextOnPage(fullName) shouldBe true
-    Helpers.isTextOnPage(nino) shouldBe true
+    val displayedNino = nino.take(2) + " " + nino.drop(2).take(2) + " " + nino.drop(4).take(2) + " " + nino.drop(6).take(2) + " " + nino.drop(8).take(1)
```
AndyHWChung: you could also do:
```scala
nino.grouped(2).mkString(" ")
```

selenium-system-test/src/test/scala/hts/utils/Helpers.scala
```diff
@@ -17,9 +17,22 @@
 package hts.utils
 
 import hts.pages.Page
+import org.openqa.selenium.support.ui.{ExpectedConditions, WebDriverWait}
 import org.openqa.selenium.WebDriver
 
 object Helpers extends Page {
 
-  def isTextOnPage(text: String)(implicit driver: WebDriver): Boolean = text.r.findAllIn(getPageContent).nonEmpty
+  def isTextOnPage(text: String)(implicit driver: WebDriver): Boolean = {
```
AndyHWChung: it'd be more informative to call the parameter `regex` rather than `text`

selenium-system-test/src/test/scala/hts/utils/Helpers.scala
```diff
@@ -17,9 +17,22 @@
...
+    val textPresent = text.r.findAllIn(getPageContent).nonEmpty
+    if (!(textPresent)) {
+      println("Text not found: " + text)
+    }
+    textPresent
+  }
+
+  def isActualUrlExpectedUrl(expectedUrl: String)(implicit driver: WebDriver): Boolean = {
```
AndyHWChung: this method never actually returns `false`. What you could probably do to make this a bit cleaner is to do something like this:
```scala
def checkIsActualUrlExpectedUrl(expectedUrl: String)(implicit driver: WebDriver): Unit = {
 val wait = new WebDriverWait(driver, 20)
  wait.until(ExpectedConditions.urlContains(expectedUrl))
 }
```

selenium-system-test/src/test/scala/hts/utils/Helpers.scala
AndyHWChung: also, when you're checking URLs you do a wait to check if the url is eventually the one you expect. When you're checking text on a page do you need to do something similar or do you want to check the condition once only?

MarcusQA: Once only for now.

AndyHWChung:
LGTM

PR #387, opened 2017-12-11
Hts 256 personal details unavailable

thebailers:
Update content on missing personal details page


conf/messages
```diff
@@ -206,12 +206,15 @@ hts.register.missinguserinfo.header=We''re missing some important details
 hts.register.missinguserinfo.header.page.title=We''re missing some important details
 hts.register.missinguserinfo.p1=We can''t create your account until we have your:
 hts.register.missinguserinfo.title2=What happens next?
-hts.register.missinguserinfo.next.step1=You''ll need to go to the form on the HM Revenue & Customs website and enter your details.
+hts.register.missinguserinfo.next.step1=You''ll need to <a href={0} target="_blank">fill in the form (it will open in a new window)</a> on the HM Revenue & Customs website.
```
AndyHWChung: the url of the link here is a parameter but later on in this file we have a hardcoded url for the exit link. Do we want urls hard coded in the messages file or in the view file? It will have the same result, but this is just for consistency. It may make more sense to have URLs in the views file rather than the messages file since URLs should be independent of language

thebailers: Updated - moved URL from the messages file to the view.

AndyHWChung:
LGTM

PR #397, opened 2017-12-13
Privacy policy fixes

fordie:


conf/messages
```diff
@@ -406,9 +406,12 @@ hts.privacy.ul2.li2=protection of public funds
 hts.privacy.h2-1=HMRC staff handling your information
 hts.privacy.p10=HMRC''s staff are trained in handling information and understand how important it is to protect personal and other sensitive information.
 hts.privacy.h2-2=Seeing what information we hold
-hts.privacy.p11=You can ask to see the personal information HMRC holds about you. To help locate the information you want and deal with your request more quickly, you should make your Data Protection Act request to the HMRC office that you''ve been dealing with. You should address your request to the ''Data Protection Officer''.  You can get more information on HMRC''s privacy policy (it will open in a new window) on the GOV.UK website.
+hts.privacy.p11.1=You can ask to see the personal information HMRC holds about you. To help locate the information you want and deal with your request more quickly, you should make your Data Protection Act request to the HMRC office that you''ve been dealing with. You should address your request to the ''Data Protection Officer''. 
+hts.privacy.p11.link=https://www.gov.uk/government/publications/data-protection-act-dpa-information-hm-revenue-and-customs-hold-about-you
```
AndyHWChung: does the link url belong here or in the view file? Is the url going to change with different languages?

fordie: Good question. I suppose there's a good chance that there's a Welsh version, but I don't know whether that has a different URL

AndyHWChung: should we stick that URL in the view file rather than the messages file then?

fordie: can do

AndyHWChung:
LGTM

PR #404, opened 2017-12-14
HTS-639 Update main heading sizeing xlarge to large

thebailers:
Global update of page heading sizing from xlarge to large


AndyHWChung:
If we had something in `helpers` that returned a header with the correct size in, if we ever had to change the size again we'd only have to change it in one place instead of 20+ places. This is probably not really important for now though - just a thought for the future

AndyHWChung:
LGTM

PR #409, opened 2017-12-14
Hts 636

Jackie1988:


test/uk/gov/hmrc/helptosavefrontend/controllers/RegisterControllerSpec.scala
```diff
@@ -67,7 +68,23 @@ class RegisterControllerSpec
...
     mockAuditor,
     fakeApplication,
-    mockPagerDuty)(
+    mockPagerDuty,
+    Configuration("enable-early-cap-check" → false))(
+    crypto, mockEmailValidation, transformer) {
+    override lazy val authConnector = mockAuthConnector
+  }
+
+  val trueEarlyCapController: RegisterController = new RegisterController(
```
AndyHWChung: do we need this `val` in scope for the whole spec, it makes the spec a bit messy. What you could do is something like this:
```scala
def newController(earlyCapCheck: Boolean): RegisterController = new RegisterController(...)
val controller = newController(false)
```
and in each spec where you need it true you can do
```
  "blah blah blah" in {
     val controller = newController(true)
     ...
  }
```

Jackie1988: done, thanks for the idea :)

PR #436, opened 2017-12-21
SSL debug logging :  verbose = true

sureshhmrc:
```hmm, just checked on QA, no ssl logs, i remember it worked last time with enabling only `handshake = true` , we can try with setting `verbose = true`  too , but remember these are not recommended on production```


AndyHWChung:
LGTM

PR #443, opened 2018-01-05
HTS-668: Verify the feedback, get-help, and privacy links appear on a…

sureshhmrc:
…ll HTS pages

### DIFF
--- a/selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
+++ b/selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
@@ -21,6 +21,8 @@ import hts.pages._
 import hts.pages.registrationPages._
 import hts.utils.ScenarioContext
 
+import scala.collection.JavaConverters._
+
 class VerifyLinksSteps extends Steps {
 
   Given("""^that users are authenticated$""") { () ⇒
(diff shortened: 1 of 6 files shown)


selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
```diff
@@ -0,0 +1,94 @@
...
+    SelectEmailPage.setAndVerifyNewEmail("newemail@mail.com")
+    Browser.checkCurrentPageIs(VerifyYourEmailPage)
+    checkForLinksThatExistOnEveryPage(SelectEmailPage)
+
+    //go back to original select email page and continue
+    SelectEmailPage.selectGGEmail()
+    //SelectEmailPage.clickContinue()
+    checkForLinksThatExistOnEveryPage(CreateAccountPage)
+
+    CreateAccountPage.createAccount()
```
AndyHWChung: do we really need to create the account at the end?

selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  }
+
+  private def checkForLinksThatExistOnEveryPage(currentPage: Page): Unit = {
+
+    //feedback link
+    Browser.isTextOnPage("BETA") shouldBe true
+    Browser.clickButtonByIdOnceClickable("feedback-link")
+    driver.navigate().back()
```
AndyHWChung: after clicking the feedback link button and before going back we should verify somehow that we are on the correct page - otherwise we could have the link redirect to an unexpected place and this test will still pass

selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    //Get Help with this page
+    Browser.isTextOnPage("Is there anything wrong with this page") shouldBe true
+    Browser.clickButtonByIdOnceClickable("get-help-action")
+
```
AndyHWChung: when you click on this button (or is it a link?) there should appear a form on the page - we should verify this before moving on to the rest of the test

selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
```diff
@@ -0,0 +1,94 @@
...
+    //Privacy policy
+    Browser.isTextOnPage("Privacy policy") shouldBe true
+    Browser.click on Browser.linkText("Privacy policy")
```
AndyHWChung: again, we should check that we're actually on the privacy policy page before we redirect to `currentPage`

selenium-system-test/src/test/scala/hts/steps/VerifyLinksSteps.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  }
+
+  private def verifyPrivacyPolicyLink(): Unit = {
+    Browser.isTextOnPage("Privacy policy") shouldBe true
+    Browser.clickLinkTextOnceClickable("Privacy policy")
+    val tabs = driver.getWindowHandles.asScala.toList
+    driver.switchTo.window(tabs(1)) //switch to privacy policy tab
+    Browser.checkCurrentPageIs(PrivacyPolicyPage)
+    driver.switchTo.window(tabs.headOption.getOrElse("missing url here")) //switch back to original page
```
AndyHWChung: since you're doing `tabs(1)` above you might as well do `tabs(0)` here - or you could even do something like:
```scala
tabs match {
  case tab1 :: tab2 :: Nil => // blah blah blah
  case ts                  => fail(s"Unexpected number of tabs: ${ts.getLength}")
}
```

AndyHWChung:
LGTM

PR #457, opened 2018-01-10
HTS-696: Added a condition to stop the script if HTS_DEP fail to start and updated health check

YordanAngelov:


services-health-check.sh
```diff
@@ -4,7 +4,7 @@ echo "*********** Running service health check ***********"
 
 SM_STATUS_OUTPUT=`sm -s`
 echo $SM_STATUS_OUTPUT
-SERVICES=(HELP_TO_SAVE_FRONTEND HELP_TO_SAVE HELP_TO_SAVE_STUB ASSETS_FRONTEND AUTH AUTH_LOGIN_API AUTH_LOGIN_STUB AUTH_DES_STUB AUTHENTICATOR CITIZEN_FRONTEND DATASTREAM GG_STUBS GG GG_AUTHENTICATION IDENTITY_VERIFICATION IDENTITY_VERIFICATION_FRONTEND IDENTITY_VERIFICATION_STUB KEYSTORE PERTAX_FRONTEND USER_DETAILS CA_FRONTEND EMAIL_VERIFICATION EMAIL_VERIFICATION_FRONTEND EMAIL_VERIFICATION_STUBS EMAIL HMRC_EMAIL_RENDERER MAILGUN_STUB)
+SERVICES=(HELP_TO_SAVE_FRONTEND, HELP_TO_SAVE, HELP_TO_SAVE_STUB, ASSETS_FRONTEND, AUTH, AUTH_LOGIN_API, AUTH_LOGIN_STUB, AUTH_DES_STUB, AUTHENTICATOR, CITIZEN_DETAILS, CITIZEN_FRONTEND, DATASTREAM, GG_STUBS, GG, GG_AUTHENTICATION, IDENTITY_VERIFICATION, IDENTITY_VERIFICATION_FRONTEND, IDENTITY_VERIFICATION_STUB, KEYSTORE, USER_DETAILS, CA_FRONTEND, EMAIL_VERIFICATION, EMAIL_VERIFICATION_FRONTEND, EMAIL_VERIFICATION_STUBS, EMAIL, HMRC_EMAIL_RENDERER, MAILGUN_STUB, CONTACT_FRONTEND, HMRCDESKPRO)
```
AndyHWChung: does this actually work? do these need to be strings like in the answer to this question here https://stackoverflow.com/questions/8880603/loop-through-an-array-of-strings-in-bash?

PR #470, opened 2018-01-24
Hts 663

thebailers:
Updates the introduction copy, sidebar alignment with the main content, and the font size of the introduction navigation.


AndyHWChung:
LGTM

PR #490, opened 2018-02-16
Reconfigured call to NS&I to talk to help-to-save-proxy

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -44,22 +42,18 @@ import scala.concurrent.Future
...
-                                    app:                         Application,
-                                    pagerDutyAlerting:           PagerDutyAlerting,
-                                    configuration:               Configuration
+class RegisterController @Inject() (val messagesApi:           MessagesApi,
+                                    val helpToSaveService:     HelpToSaveService,
+                                    val sessionCacheConnector: SessionCacheConnector,
+                                    frontendAuthConnector:     FrontendAuthConnector,
+                                    metrics:                   Metrics,
+                                    app:                       Application,
+                                    pagerDutyAlerting:         PagerDutyAlerting,
```
sureshhmrc: i guess we dont need `pagerDutyAlerting` here anymore ?

AndyHWChung: no we don't need it - I'll remove it

PR #493, opened 2018-02-27
HTS-854: Keep IV session data in key-store throughout session

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/connectors/SessionCacheConnector.scala
```diff
@@ -54,26 +55,37 @@ class SessionCacheConnectorImpl @Inject() (val http: WSHttp, metrics: Metrics)
...
+                        pendingEmail           = newSession.pendingEmail
+          )
+        )
+
+        cache[HTSSession](sessionKey, sessionToStore)(writes, hc, ec).map { cacheMap ⇒
+          val _ = timerContext.stop()
+          cacheMap
+        }
+      }
+    ).leftMap {
```
AndyHWChung: doing `leftMap` here doesn't handle the case when the `Future` fails, we need a `recover` block for that. You need something like this:
```scala
  def f(newSession: HTSSession, oldSession: Option[HTSSession])(implicit writes: Writes[HTSSession], hc: HeaderCarrier, ec: ExecutionContext) = {
    val timerContext = metrics.keystoreWriteTimer.time()
    val sessionToStore = oldSession.fold(
      newSession
    )(existing ⇒
      existing.copy(eligibilityCheckResult = newSession.eligibilityCheckResult,
        confirmedEmail         = newSession.confirmedEmail,
        pendingEmail           = newSession.pendingEmail
      )
    )

    cache[HTSSession](sessionKey, sessionToStore).map { cacheMap ⇒
      val _ = timerContext.stop()
      cacheMap
    }.recover{
      case NonFatal(e) ⇒
        val _ = timerContext.stop()
      metrics.keystoreWriteErrorCounter.inc()
      Left(s"Oh no, something went wrong: ${e.getMessage}")
    }
  }
  
  def put(newSession: HTSSession)(implicit writes: Writes[HTSSession], hc: HeaderCarrier, ec: ExecutionContext): Result[CacheMap] =
    for{
      oldSession ← get
      result ← f(newSession, oldSession)
    } yield result
```

AndyHWChung: `leftMap` is normally used to map the left type of the either into something else rather than performing side-effecting functions

AndyHWChung:
LGTM

PR #494, opened 2018-02-28
HTS-836: Added missing email validation rules

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/views/helpers/email_form_validation_error_message.scala.html
```diff
@@ -32,6 +32,12 @@
         "hts.email.error.total-too-long"
     } else if(form.emailHasNoAtSymbol(key)) {
         "hts.email.error.no-at-symbol"
+    } else if(form.emailHasNoDotSymbol(key)) {
+        "hts.email.error.no-dot-symbol"
```
AndyHWChung: we need to double check with Jean what these messages should be

AndyHWChung:
LGTM

PR #495, opened 2018-03-02
Hts 762 - UC copy changes

thebailers:


thebailers:
Work merged to uc branch for release with UC changes in April

PR #502, opened 2018-03-06
HTS-764

sophie-still:
Technical error scenario now passing locally


selenium-system-test/src/test/scala/hts/utils/ScenarioContext.scala
```diff
@@ -104,12 +104,25 @@ private[utils] trait NINOGenerator {
     prefixes.toVector(rnd.nextInt(prefixes.size))
   }
 
+  def generateHTTPErrorCodePrefix(code: Int): String = {
+    val codePrefix = code.toString
+    val prefix = "ES" + codePrefix
+    val rnd = new Random
```
MarcusQA: Is the rnd variable used, Sophie?

PR #518, opened 2018-03-22
hts-856 adds UC & wtc not eligible copy

thebailers:

PR #535, opened 2018-04-23
Hts 762 - UC copy changes (#497)

thebailers:
hts-856 adds UC & wtc not eligible copy (#518)
HTS-856 - fix think you are eligible link (#519)


AndyHWChung:
LGTM

PR #537, opened 2018-04-24
 HTS-986: Use defaultURLs during IV if session data is not found

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/controllers/IvController.scala
```diff
@@ -190,10 +194,18 @@ class IvController @Inject() (val sessionCacheConnector: SessionCacheConnector,
...
-        internalServerError()
-      }(f)
+      mayBeSession ⇒
+        mayBeSession.fold {
+          logger.warn(s"no session found for user in the keystore, redirecting to $defaultUrl")
+          SeeOther(defaultUrl)
+        } { session ⇒
+          url(session).fold {
+            logger.warn("session exists in the keystore but required information is not found")
+            internalServerError()
```
AndyHWChung: in this case would we not do the same as above and do `SeeOther(defaultUrl)`?

AndyHWChung:
LGTM

PR #567, opened 2018-06-12
hts-1043 adds dynamic close account page

thebailers:


app/uk/gov/hmrc/helptosavefrontend/controllers/AccountHolderController.scala
```diff
@@ -151,10 +151,14 @@ class AccountHolderController @Inject() (val helpToSaveService:          HelpToS
...
             e ⇒ {
               logger.warn(s"error retrieving Account details from NS&I, error = $e", htsContext.nino)
-              internalServerError()
+              Ok(views.html.close_account_are_you_sure(None))
             },
-            {
-              account ⇒ Ok(views.html.close_account_are_you_sure(account))
+            { account ⇒
+              if (account.isClosed) {
+                SeeOther(appConfig.nsiManageAccountUrl)
```
Jackie1988: So if there account is already closed, we're just gonna direct the user to nsandi.com with no comms to say their account has already been closed?

thebailers: The user will be redirected to their account homepage which shows that the account has been closed.

PR #568, opened 2018-06-12
HTS-876: Added scenarios for verifying email address

MarcusQA:


selenium-system-test/src/test/scala/hts/steps/Steps.scala
```diff
@@ -82,6 +83,11 @@ private[steps] trait Steps extends ScalaDsl with EN with Matchers {
     }
   }
 
+  def createAccountUsingGGEmail(): Unit = {
```
AndyHWChung: this doesn't quite belong here - this is specific to the steps we're implementing rather than steps in general - maybe it could go in the `steps` package object or somewhere similar?

PR #570, opened 2018-06-14
HTS-1093: Pass eligibility reason to BE during create-account

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/models/register/CreateAccountRequest.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.helptosavefrontend.models.register
+
+import play.api.libs.json.{Format, Json}
+import uk.gov.hmrc.helptosavefrontend.models.userinfo.NSIUserInfo
+
+case class CreateAccountRequest(userInfo: NSIUserInfo, eligibilityReason: Option[Int])
```
AndyHWChung: I think in all cases in the digital FE we're gonna have the `eligibilityReason` when creating accounts - this can safely be an `Int` instead of an `Option[Int]`

AndyHWChung:
LGTM

PR #590, opened 2018-07-11
hts-1115 prefix page titles with Error on form validation error

thebailers:


app/uk/gov/hmrc/helptosavefrontend/views/email/cannot_change_email.scala.html
```diff
@@ -20,13 +20,14 @@
 @import views.html.helper.CSRF
 @import uk.gov.hmrc.helptosavefrontend.forms.Continue
 @import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig
+@import uk.gov.hmrc.helptosavefrontend.views.html.helpers.error_prefix
```
AndyHWChung: in the implementation of `error_prefix`:
```
@(form: Form[_])(implicit messages: Messages)
  @if(form.hasErrors){
    @Messages("hts.global.error.prefix")
  }
```
is it safer to have an else case? e.g.
```
@(form: Form[_])(implicit messages: Messages)
  @if(form.hasErrors){
    @Messages("hts.global.error.prefix")
  } else {
    ""
  }
```
or whatever the syntax would be? From a types point of view this is preferable so that the inferred type is `String` rather than `Any`

thebailers: Thanks - have updated the helper.

AndyHWChung:
LGTM

PR #592, opened 2018-07-13
HTS-1084: Do not start email verification for DE users if they have a…

sureshhmrc:
… valid existing GG

### DIFF
--- a/app/uk/gov/hmrc/helptosavefrontend/controllers/EmailController.scala
+++ b/app/uk/gov/hmrc/helptosavefrontend/controllers/EmailController.scala
@@ -242,31 +242,32 @@ class EmailController @Inject() (val helpToSaveService:          HelpToSaveServi
     }.merge
   }
 
-  def confirmEmail(email: String): Action[AnyContent] = authorisedForHtsWithNINO { implicit request ⇒ implicit htsContext ⇒
+  def confirmEmail(email: String): Action[AnyContent] = authorisedForHtsWithInfo { implicit request ⇒ implicit htsContext ⇒
     val nino = htsContext.nino
+    val decryptedEmailEither = EitherT.fromEither[Future](decryptEmail(email))
 
-      def doUpdate(hTSSession: HTSSession)(ifSuccess: ⇒ Result) = {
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/helptosavefrontend/controllers/EmailController.scala
```diff
@@ -242,31 +242,32 @@ class EmailController @Inject() (val helpToSaveService:          HelpToSaveServi
...
+        result.fold(
           { e ⇒
             logger.warn(s"Could not write confirmed email: $e", nino)
-            internalServerError()
+            toFuture(internalServerError())
           }, { _ ⇒
             ifSuccess
           }
-        )
+        ).flatMap(identity)
```
AndyHWChung: Is a `fold(..).flatMap(identity)` necessary here. You could write it as part of the for-comprehension above:
```scala
val result = for {
  e ← decryptedEmailEither
  _ ← sessionCacheConnector.put(hTSSession.copy(confirmedEmail = Some(e)))
  _ ← helpToSaveService.storeConfirmedEmail()
 r   ← EitherT.liftF(ifSuccess)
  } yield r
```

app/uk/gov/hmrc/helptosavefrontend/controllers/EmailController.scala
```diff
@@ -276,9 +277,34 @@ class EmailController @Inject() (val helpToSaveService:          HelpToSaveServi
       def ifDE = {
         _: HTSSession ⇒
           {
-            doUpdate(HTSSession(None, None, None, false))(
-              SeeOther(routes.EmailController.verifyEmail().url)
-            )
+            doUpdate(HTSSession(None, None, None, false)) {
+              htsContext.userDetails.fold[Future[Result]](
```
AndyHWChung: this is very similar to the behaviour under `confirmEmail` - would it be possible to return `SeeOther(routes.EmailController.confirmEmail()` here to re-use some of the logic there?

AndyHWChung:
LGTM

PR #618, opened 2018-08-14
hts-1238 update exit link text and destination on failed iv locked out

thebailers:


app/uk/gov/hmrc/helptosavefrontend/views/iv/locked_out.scala.html
```diff
@@ -28,6 +28,6 @@ <h1 class="heading-large">@Messages("hts.iv.locked_out.title.h1")</h1>
   <p>@Messages("hts.iv.locked_out.p1")</p>
   <p>@Html(Messages("hts.iv.locked_out.p2"))</p>
   <section class="subsection--wide">
-    <p><a href=@routes.IntroductionController.getAboutHelpToSave().url>@Messages("hts.exit_to_about_hts")</a></p>
+    <p><a href="https://www.gov.uk">@Messages("hts.global.exit.to.gov.uk")</a></p>
```
AndyHWChung: Are we happy to go to the gov.uk homepage while we are waiting for the help to save gov uk pages?

thebailers: I have spoken with Jean and Nick and they have confirmed this is fine

AndyHWChung: it might be worth sticking this in `AppConfig` then. If we make other links go to gov.uk we can reference the `val` in `AppConfig` and when we update the link location we'll then only have to do a config change to make it happen

PR #623, opened 2018-08-16
HTS-1237 Updated content on no account page

Jackie1988:


app/uk/gov/hmrc/helptosavefrontend/views/core/confirm_check_eligibility.scala.html
```diff
@@ -26,11 +26,18 @@
 ) {
   <h1 class="heading-large">@Messages("hts.no-account.title.h1")</h1>
 
-  <p class="lede">@Html(Messages("hts.no-account.p1"))</p>
+  <p>@Html(Messages("hts.no-account.p1"))</p>
```
AndyHWChung: is there a reason you got rid of the `lede` class here?

AndyHWChung: c.f. https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/Lead_paragraph.html

Jackie1988: Because the size of the text in the lede class doesn't match the new content

Jackie1988: I think we should check this with Nick and Jean

Jackie1988: I have just checked this with Nick and he has asked to put it back, so raising another PR

PR #627, opened 2018-08-21
HTS-1251 : Added Bank Account Details Page

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/forms/BankDetails.scala
```diff
@@ -0,0 +1,35 @@
...
+import play.api.data.Form
+import play.api.data.Forms._
+import play.api.libs.json.{Format, Json}
+
+object BankDetails {
+  def giveBankDetailsForm(): Form[BankDetails] = Form(
+    mapping("sortCode" -> nonEmptyText(6, 6),
+      "accountNumber" -> nonEmptyText(maxLength = 10),
+      "rollNumber" -> optional(text),
+      "accountName" -> nonEmptyText(maxLength = 50)
```
sureshhmrc: added some dummy validation for now,  do we have an agreed field lengths for each field above ?

app/uk/gov/hmrc/helptosavefrontend/views/register/bank_account_details.scala.html
```diff
@@ -0,0 +1,65 @@
...
+@import uk.gov.hmrc.helptosavefrontend.views.helpers.errorPrefix
+@import uk.gov.hmrc.helptosavefrontend.controllers.routes
+@import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig
+
+@(form: Form[BankDetails], referrer: String)(implicit htsContext: HtsContext, request: Request[_], messages: Messages, appConfig: FrontendAppConfig)
+
+@uk.gov.hmrc.helptosavefrontend.views.html.main_template(title = "Enter Bank Details", bodyClasses = None) {
+@helpers.form(routes.BankAccountController.submitBankDetails(), 'id -> "bankDetailsForm") {
+
+<a href="@referrer">Back</a>
```
AndyHWChung: not sure we'll need the referrer here - should be able to get the previous page via the browser history, e.g. https://www.w3schools.com/jsref/met_his_back.asp

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -0,0 +1,93 @@
...
+                                       val authConnector:         AuthConnector,
+                                       val metrics:               Metrics)(implicit override val messagesApi: MessagesApi,
+                                                                           val transformer:       NINOLogMessageTransformer,
+                                                                           val frontendAppConfig: FrontendAppConfig,
+                                                                           val config:            Configuration,
+                                                                           val env:               Environment)
+
+  extends BaseController with HelpToSaveAuth with EnrolmentCheckBehaviour with SessionBehaviour {
+
+  def getBankDetailsPage(): Action[AnyContent] = authorisedForHtsWithNINO { implicit request ⇒ implicit htsContext ⇒
```
AndyHWChung: do we need to retrieve the NINO here? Could we get away with `authorisedForHts`?

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -0,0 +1,93 @@
...
+
+                    ineligibilityType.fold {
+                      logger.warn(s"Could not parse ineligibility reason when storing bank details: $ineligibleReason", htsContext.nino)
+                      toFuture(internalServerError())
+                    } { i ⇒
+                      toFuture(Ok(views.html.register.not_eligible(i)))
+                    }
+                  },
+                  { _ ⇒
+                    sessionCacheConnector.put(session.copy(bankDetails = Some(bankDetails)))
```
AndyHWChung: would a `.fold` not work here?

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -129,6 +128,10 @@ class RegisterController @Inject() (val helpToSaveService:     HelpToSaveService
     }
   }(redirectOnLoginURL = routes.RegisterController.getCreateAccountPage().url)
 
+  def checkYourDetails(): Action[AnyContent] = authorisedForHtsWithInfo { implicit request ⇒ implicit htsContext ⇒
```
AndyHWChung: would we need `authorisedForHtsWithInfo` here? Would we just not grab the user info from session if the person is eligible?

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
sureshhmrc: `authorisedForHts`  doesn't demand CL200 which is annoying , but here we expect CL200, which we can fix/refactor in `HelpToSaveAuth` later

AndyHWChung: oh, I think that's because it's being used for the introduction pages where we don't need CL200. Since those pages are migrating to gov.uk soon, we'll be able to make authorisedForHts require CL200 after that. For now, using `authorisedForHtsWithNINO` should be fine - the extra retrieving of the NINO isn't going to break anything

AndyHWChung:
LGTM

PR #631, opened 2018-08-24
Hts 1277

Jackie1988:


sureshhmrc:
did you consider the new url for conditions page ?
ticket talks about `/agree-conditions-to-create-account`

Jackie1988:
I've updated the /create-account URL to /agree-conditions-to-create-account

PR #648, opened 2018-09-04
HTS-1237 Updated content on no account page (#623)

AndyHWChung:
HTS-1237 updated exit link on confirm_check_eligibility page (#625)


thebailers:
Looks good.

PR #652, opened 2018-09-06
Handle JSON parsing of SubmissionFailure correctly

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/services/HelpToSaveService.scala
```diff
@@ -126,7 +126,7 @@ class HelpToSaveServiceImpl @Inject() (helpToSaveConnector: HelpToSaveConnector)
     )
 
   private def handleError(response: HttpResponse): SubmissionFailure = {
-    response.parseJSON[SubmissionFailure](Some("error")) match {
```
sureshhmrc: incase of submisson failures the error json from proxy doesn't contain the path `error`, instead it straightaway has the SubmissionFailure json something like 
```{"errorMessageId":"","errorMessage":"Invalid data found in request","errorDetail":"The following fields were either invalid or missing: [obj.nbaDetails.sortCode]"}```

AndyHWChung:
LGTM

PR #666, opened 2018-09-11
HTS-1247 - implemented basic bank account validation

AndyHWChung:


app/uk/gov/hmrc/helptosavefrontend/config/FrontendAppConfig.scala
```diff
@@ -99,4 +99,14 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
...
   val betaFeedbackUrlNoAuth: String = s"$contactBaseUrl/contact/beta-feedback-unauthenticated?service=$contactFormServiceIdentifier"
   val govUkExitURL: String = getString("gov-uk.url")
+
+  object BankDetailsValidation {
+    val sortCodeLength: Int = getInt("bank-details-validation.sort-code.length")
+    val accountNumberLength: Int = getInt("bank-details-validation.account-number.length")
+    val rollNumberMinLength: Int = getInt("bank-details-validation.roll-number.min-length")
+    val rollNumberMaxLength: Int = getInt("bank-details-validation.roll-number.max-length")
+    val accountNameMinLength: Int = getInt("bank-details-validation.account-name.min-length")
+    val accountNameMaxLength: Int = getInt("bank-details-validation.account-name.max-length")
```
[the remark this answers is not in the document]
AndyHWChung: if we ever need to change those values I'd rather it be a config change rather than a code change. It is unlikely to change, but I think it's best to be prepared just in case.

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
```diff
@@ -123,17 +123,17 @@ object BankDetailsValidation {
 
   private[forms] object ErrorMessages {
 
-    val sortCodeIncorrectFormat = "sort-code-incorrect-format"
+    val sortCodeIncorrectFormat = "sort_code_incorrect-_format"
```
sureshhmrc: missed one hyphen `-_` ?

AndyHWChung: whoops, missed that one! Fixed now

app/uk/gov/hmrc/helptosavefrontend/config/FrontendAppConfig.scala
```diff
@@ -99,4 +99,14 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   val reportAProblemNonJSUrl: String = s"$contactBaseUrl/contact/problem_reports_nonjs?service=$contactFormServiceIdentifier"
   val betaFeedbackUrlNoAuth: String = s"$contactBaseUrl/contact/beta-feedback-unauthenticated?service=$contactFormServiceIdentifier"
   val govUkExitURL: String = getString("gov-uk.url")
+
+  object BankDetailsConfig {
+    val sortCodeLength: Int = getInt("bank-details-validation.sort-code.length")
```
sureshhmrc: also isn't `sort-code` just word , like `sortCode` and so hyphen is not required for it , but this needs to change at lot of places, so we can live with it for now 👍

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
```diff
@@ -0,0 +1,172 @@
...
+
+    val accountNumberIncorrectFormat = "account_number_incorrect_format"
+
+    val rollNumberTooShort = "roll_number_too_short"
+
+    val rollNumberTooLong = "roll_number_too_long"
+
+    val accountNameTooShort = "account_name_too_short"
+
+    val accountNameTooLong = "account_name_too_long"
```
sureshhmrc: also i was thinking these error key strings should appear the same way in messages.conf too, but i see in its `sort-code.incorrect-format` in the messages file, cant we keep the same key at both places ? 
keeping the same key at both places help in searching/finding for those occurrences in the whole codebase ?

AndyHWChung: it doesn't really matter what the values of these strings are in reality - all they do is it facilitate the interface between the BE and the FE view. They're independent of what's in the messages file. The values themselves don't drive any logic except in the `FormOps` implicit class

PR #667, opened 2018-09-11
HTS-1294: Update Selenium tests to reflect new registration journey

MarcusQA:


conf/messages
```diff
@@ -273,7 +273,7 @@ hts.register.bank-details.rollNumber.label=Roll number
 hts.register.bank-details.rollNumber.hint=Building societies only
 
 # Eligible
-hts.register.you-are-eligible.title.h1=You’re eligible for a Help to Save account.
+hts.register.you-are-eligible.title.h1=You’re eligible for a Help to Save account
```
AndyHWChung: need to use smart quotes here - look for `&rsquo;` in this file

selenium-system-test/src/test/scala/hts/pages/AuthorityWizardPage.scala
```diff
@@ -98,7 +98,7 @@ object AuthorityWizardPage extends Page {
   def setNino(nino: String)(implicit driver: WebDriver): Unit =
     Browser.find(Browser.name("nino")).foreach { element ⇒
       element.underlying.sendKeys(nino)
-      println("NINO entered on auth wizard: " + nino)
+      ScenarioContext.currentNINO
```
AndyHWChung: this line isn't doing anything - the function returns `Unit` and this line returns a `String`

selenium-system-test/src/test/scala/hts/pages/BankDetailsPage.scala
```diff
@@ -0,0 +1,50 @@
...
+
+object BankDetailsPage extends Page {
+
+  val expectedURL: String = s"${Configuration.host}/help-to-save/enter-uk-bank-details"
+
+  override val expectedPageHeader: Option[String] = Some("Which UK bank account do you want us to pay your bonuses and withdrawals into?")
+
+  override val expectedPageTitle: Option[String] = Some("Which UK bank account do you want us to pay your bonuses and withdrawals into?")
+
+  def enterValidDetails()(implicit driver: WebDriver): Unit = {
```
AndyHWChung: it would be more general to have a `enterDetails` function instead with the details being passed in so that the function can be reused to pass in invalid details if we ever need to. The `vals` in the `ValidBankDetails` object would then move to the calling code

AndyHWChung:
LGTM

PR #673, opened 2018-09-13
HTS-1250: Added BARS validation

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/services/BarsService.scala
```diff
@@ -0,0 +1,71 @@
...
+        response.status match {
+          case Status.OK ⇒
+            (response.json \ "accountNumberWithSortCodeIsValid").asOpt[Boolean] match {
+              case Some(result) ⇒ Right(result)
+              case None ⇒
+                logger.warn(s"error parsing the response from bars check, trackingId = $trackingId,  body = ${response.body}")
+                Left(s"error parsing the response json from bars check")
+            }
+          case other: Int ⇒
+            //Do we need pager duty alert here ?
```
AndyHWChung: it'd be safer to put a pager duty alert here - we should probably have an event when we couldn't parse the JSON response as well

conf/messages
```diff
@@ -268,6 +268,8 @@ hts.register.bank-details.accountNumber.label=Account number
 hts.register.bank-details.accountName.label=Account name
 hts.register.bank-details.rollNumber.label=Roll number
 hts.register.bank-details.rollNumber.hint=Building societies only
+hts.bank_details.check_your_sortcode_is_correct = Check your sort code is correct
```
AndyHWChung: Does the leading space after the `=` cause an extra space in the UI?

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -68,17 +69,30 @@ class BankAccountController @Inject() (val helpToSaveService:     HelpToSaveServ
...
+                      logger.warn(s"Could not update session with bank details: $error")
+                      internalServerError()
+                    },
+                    _ ⇒ SeeOther(routes.RegisterController.checkDetails().url)
+                  )
+            }
+
+          case Right(false) ⇒
+            val formWithErrors = BankDetails.giveBankDetailsForm().fill(bankDetails)
+              .withError("sortCodeBars", "check_your_sortcode_is_correct")
```
AndyHWChung: it'd be more flexible for the future if the message key wasn't hard coded here. If we ever change the message keys then someone might forget to update it here.

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -68,17 +69,30 @@ class BankAccountController @Inject() (val helpToSaveService:     HelpToSaveServ
...
-                error ⇒ {
-                  logger.warn(s"Could not update session with bank details: $error")
-                  internalServerError()
-                },
-                _ ⇒ SeeOther(routes.RegisterController.checkDetails().url)
-              )
-        }
+        barsService.validate(bankDetails).map[Future[Result]] {
+          case Right(true) ⇒
+            checkIfAlreadyEnrolledAndDoneEligibilityChecks(htsContext.nino) {
```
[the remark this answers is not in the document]
sureshhmrc: but if users reached this point means mostly they have gone through the enrolment and eligibility checks. 
if we do them before and incase the bank details validation failed then  we are wasting two calls : enrolment and eligibility

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -68,17 +69,30 @@ class BankAccountController @Inject() (val helpToSaveService:     HelpToSaveServ
...
+              session ⇒
+                val sortCodeWithHyphens = bankDetails.sortCode.grouped(2).mkString("-")
+                sessionCacheConnector.put(session.copy(bankDetails = Some(bankDetails.copy(sortCode = sortCodeWithHyphens))))
```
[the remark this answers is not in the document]
sureshhmrc: is this not an overkill for sortcode ? because by this point we are sure that the input has 6 digits in it without any spaces so `.grouped(2).mkString("-")` looks easier ?

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
sureshhmrc: yeah, any better approach for this ?

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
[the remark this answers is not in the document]
AndyHWChung: the type doesn't guarantee that that's the case - the type is `String` which can be anything. It feels safer to not rely on the String to be a particular format as assumptions can be wrong

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
AndyHWChung: perhaps you may be able to add something to `FormOps` in `BankDetailsValidation` as that interface is being used already for the same purpose

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetails.scala
```diff
@@ -32,4 +32,24 @@ object BankDetails {
...
+object SortCode {
+  def apply(digits: Seq[Int]): Option[SortCode] = digits.toList match {
+    case d1 :: d2 :: d3 :: d4 :: d5 :: d6 :: Nil ⇒ Some(SortCode(d1, d2, d3, d4, d5, d6))
+    case _                                       ⇒ None
+  }
+
+  implicit val writes: Writes[SortCode] = Json.writes[SortCode]
+
+  implicit val reads: Reads[SortCode] = Reads[SortCode](s ⇒ {
+    SortCode(s.as[String].map(_.asDigit)) match {
```
AndyHWChung: the bit insde `SortCode(...)` may throw - it'd be safer to do:
```scala
Try(s.as[String].map(_.asDigit)).toOption.flatMap(SortCode) match {
...
}

AndyHWChung:
LGTM

PR #678, opened 2018-09-14
Hts 1313 combine check details and confirm details screen

thebailers:


app/uk/gov/hmrc/helptosavefrontend/views/register/create_account.scala.html
```diff
@@ -14,55 +14,83 @@
...
 
-  <h1 class="heading-large">
-    @title
-  </h1>
+  <p class="lede">@Html(Messages("hts.cya.intro"))</p>
 
-  <p class="lede">
-    @Messages("hts.register.create_account.p1")
-  <p>
+  @summary_section_head(heading = Messages("hts.cya.your-details"))
```
AndyHWChung: is `cya` check your answers? They're not really checking their answers if it is...

thebailers: It is, the form pattern is referred to as check you answers, but I can update this.

AndyHWChung:
LGTM

PR #692, opened 2018-09-24
Pointing FE at Help to Save BARS stub

MarcusQA:


selenium-system-test/src/test/scala/hts/utils/TestBankDetails.scala
```diff
@@ -26,8 +26,8 @@ object TestBankDetails {
   val ValidBankDetails: TestBankDetails = TestBankDetails(
     //For use with BARS stub
```
AndyHWChung: is this comment still valid?

AndyHWChung:
LGTM

PR #710, opened 2018-10-08
HTS-1399: Revisit RollNumber validation

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
```diff
@@ -80,18 +84,14 @@ class BankDetailsValidation @Inject() (configuration: FrontendAppConfig) {
...
         data.get(key)
-          .map(_.cleanupSpecialCharacters.removeAllSpaces)
           .filter(_.nonEmpty)
-          .fold[ValidOrErrorStrings[Option[String]]](Valid(None)){ s ⇒
-            if (s.length < rollNumberMinLength) {
-              invalid(ErrorMessages.rollNumberTooShort)
-            } else if (s.length > rollNumberMaxLength) {
-              invalid(ErrorMessages.rollNumberTooLong)
-            } else {
+          .fold[ValidOrErrorStrings[Option[String]]](Valid(None)) { s ⇒
```
sureshhmrc: don't we need to do the same validation in the `help-to-save-api` as well ?

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
```diff
@@ -16,23 +16,27 @@
...
 import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig
-import uk.gov.hmrc.helptosavefrontend.util.Validation.{ValidOrErrorStrings, invalid, validatedFromBoolean}
 import uk.gov.hmrc.helptosavefrontend.forms.BankDetailsValidation.{ErrorMessages, StringOps}
+import uk.gov.hmrc.helptosavefrontend.util.Validation.{ValidOrErrorStrings, invalid, validatedFromBoolean}
 
 class BankDetailsValidation @Inject() (configuration: FrontendAppConfig) {
 
   import configuration.BankDetailsConfig._
 
+  private val rollNoRegex: String ⇒ Matcher = (s"^([0-9a-zA-Z\\\\/\\\\.-]{$rollNumberMinLength,$rollNumberMaxLength})" + "$").r.pattern.matcher _
```
Jackie1988: How sure are you about this regex? especially the ```\\\\/\\\\``` bit?

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
AndyHWChung: the validation in the proxy should be enough for the API journey - we have it in the FE since we can stop people in the journey before attempting to create an account if their bank details are wrong

test/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidationSpec.scala
```diff
@@ -177,16 +177,12 @@ class BankDetailsValidationSpec extends TestSupport {
...
+          testRollNumber(Some("length18stringonly"))(Right(Some("length18stringonly")))
         }
 
-        "contains trailing and leading spaces" in {
-          testRollNumber(Some("  ab1   "))(Right(Some("ab1")))
-        }
-
-        "contains control characters" in {
-          testRollNumber(Some("ab1\n\r\t"))(Right(Some("ab1")))
+        "which contain only dots, hyphens and forward slashes" in {
```
AndyHWChung: it'd be good to have a test which validates against special characters which are not `.`, `/` or `-`

sureshhmrc: we have a test case against special chars 
```
        "containing characters not allowed as per regex" in {
          testRollNumber(Some("a_b$c%&^@!"))(Left(Set(ErrorMessages.rollNumberInvalid)))
        }```

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
sureshhmrc: my Intellij automatically adding the extra `\\` , let me check the atos spec

test/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidationSpec.scala
AndyHWChung: oh yeah, sorry didn't see that

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
Jackie1988: ok

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
sureshhmrc: i checked in the  https://regex101.com/ and i have a feeling that `^([0-9a-zA-Z\/.-]{4, 18})$` is the correct one to use 
but the spec has `\\/\\`

app/uk/gov/hmrc/helptosavefrontend/forms/BankDetailsValidation.scala
sureshhmrc: i checked in the https://regex101.com/ and i have a feeling that ^([0-9a-zA-Z\/.-]{4, 18})$ is the correct one to use
but the spec has \\/\\

PR #724, opened 2018-10-24
hts-1347 updates content on account created screen

thebailers:


app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
```diff
@@ -166,14 +170,20 @@ class RegisterController @Inject() (val helpToSaveService:     HelpToSaveService
     result.fold({ e ⇒
       logger.warn(s"Could not get enrolment status or session: $e")
       internalServerError()
-    }, {
-      _.flatMap(_.accountNumber).fold(SeeOther(routes.EligibilityCheckController.getCheckEligibility().url)){
+    }, { session ⇒
+      session.flatMap(_.accountNumber).fold(SeeOther(routes.EligibilityCheckController.getCheckEligibility().url)){
         accountNumber ⇒
-          Ok(views.html.register.account_created(accountNumber))
+          val email = session.flatMap(_.confirmedEmail).fold("")(_.toString)
```
AndyHWChung: we probably don't want to show the page with an empty email. If there is no email the session is in a bad state - we're redirecting to check eligibility above so it'd make sense to do the same here. It'd make sense to have something like:
```scala
val x: Option[String,String] = for{
   s <- session
   a <- s.accountNumber
   e <- s.email
  } yield a -> e

x.fold(...)(...)
```

app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -20,7 +20,7 @@
 @import uk.gov.hmrc.helptosavefrontend.views.html.helpers.ga_submit_button
 @import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig
 
-@(accountNumber: String)(implicit htsContext: HtsContext, request: Request[_], messages: Messages, appConfig: FrontendAppConfig)
+@(accountNumber: String, email: String, endOfMonthDate: String)(implicit htsContext: HtsContext, request: Request[_], messages: Messages, appConfig: FrontendAppConfig)
```
AndyHWChung: it'd be more exact for the type of `endOfMonthDate` to be `LocalDate` rather than `String` - the FE view is then in charge of how to format that date to display

AndyHWChung: but also, the end of month date shouldn't be included as part of this ticket - (see the updated comments on the ticket), so this needs to be taken out anyway for now

Jackie1988: ok this has been removed for now

app/uk/gov/hmrc/helptosavefrontend/controllers/RegisterController.scala
Jackie1988: ok thank you, I have updated the PR

AndyHWChung:
LGTM

PR #736, opened 2018-11-06
HTS-1474: Added mongo-cacheing library and removed key-store

sureshhmrc:


conf/application.conf
```diff
@@ -185,6 +178,11 @@ microservice {
   }
 }
 
+mongodb {
+  uri = "mongodb://localhost:27017/help-to-save"
+  session.expireAfter = 1 hour
```
AndyHWChung: is this the expiry time we currently have in prod?

app/uk/gov/hmrc/helptosavefrontend/repo/SessionStore.scala
```diff
@@ -0,0 +1,139 @@
...
+      }
+
+    for {
+      oldSession ← get
+      result ← EitherT(doUpdate(newSession, oldSession))
+    } yield result
+
+  }
+
+  private[repo] def doFindById(id: Id) =
```
AndyHWChung: this function and `doCreateOrUpdate` are not that useful to have - when reading the code above where they're used, we have to scroll down to see what is happening inside the function only to see that it's only one line of code

app/uk/gov/hmrc/helptosavefrontend/repo/SessionStore.scala
```diff
@@ -0,0 +1,139 @@
...
+
+        }.recover {
+          case e ⇒
+            val _ = timerContext.stop()
+            metrics.sessionReadErrorCounter.inc()
+            Left(e.getMessage)
+        }
+
+      case None ⇒
+        // is this correct to return None if no session found in the HeaderCarrier ? or should this be an Left(error) ?
```
AndyHWChung: realistically we shouldn't expect to get to this stage since we shouldn't be looking at session data until we're sure that someone is logged in. After we know they're logged in they should have a session ID. Because of this we should probably be returning an error here since this would be an unexpected event we'd want to investigate

app/uk/gov/hmrc/helptosavefrontend/repo/SessionStore.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  private val expireAfterSeconds = appConfig.mongoSessionExpireAfter.toSeconds
+
+  private lazy val cacheRepository = new CacheMongoRepository("sessions", expireAfterSeconds)(mongo.mongoConnector.db, ec)
+
+  private type EitherStringOr[A] = Either[String, A]
+
+  override def get(implicit reads: Reads[HTSSession], hc: HeaderCarrier): Result[Option[HTSSession]] = {
+
+    val timerContext = metrics.sessionReadTimer.time()
```
AndyHWChung: this timer should start just before `doFindById` below - if there is no session id there is no need to run the timer

app/uk/gov/hmrc/helptosavefrontend/repo/SessionStore.scala
```diff
@@ -0,0 +1,139 @@
...
+      case None ⇒
+        // is this correct to return None if no session found in the HeaderCarrier ? or should this be an Left(error) ?
+        Right(None)
+    })
+  }
+
+  override def store(newSession: HTSSession)(implicit writes: Writes[HTSSession], hc: HeaderCarrier): Result[Unit] = {
+
+      def doUpdate(newSession: HTSSession, oldSession: Option[HTSSession]): Future[Either[String, Unit]] = {
+        val timerContext = metrics.sessionStoreWriteTimer.time()
```
AndyHWChung: same thing with this timer here

app/uk/gov/hmrc/helptosavefrontend/repo/SessionStore.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  def store(body: HTSSession)(implicit writes: Writes[HTSSession], hc: HeaderCarrier): Result[Unit]
+}
+
+@Singleton
+class SessionStoreImpl @Inject() (mongo: ReactiveMongoComponent, metrics: Metrics)(implicit appConfig: FrontendAppConfig, ec: ExecutionContext) extends SessionStore {
+
+  private val expireAfterSeconds = appConfig.mongoSessionExpireAfter.toSeconds
+
+  private lazy val cacheRepository = new CacheMongoRepository("sessions", expireAfterSeconds)(mongo.mongoConnector.db, ec)
```
AndyHWChung: does this need to be lazy? Is it because of the unit tests that this needs to be lazy?

test/uk/gov/hmrc/helptosavefrontend/repo/SessionStoreSpec.scala
```diff
@@ -89,28 +89,30 @@ class SessionStoreSpec extends TestSupport with MongoSupport with ScalaFutures w
...
-        result.value.futureValue should be(Right(()))
-
-        eventually {
-          val getResult = sessionStore.get(htsSessionReads, hc)
-          getResult.value.futureValue should be(Right(None))
-        }
-
-      }
-    }
+    //    "be able to delete entries automatically after TTL period is expired" in new TestApparatus {
```
AndyHWChung: is there a reason we don't want to leave this uncommented out? Is it the time it takes for the unit test?

conf/application.conf
sureshhmrc: yes, they have 60 min in prod : 
https://github.com/hmrc/app-config-production/blob/master/key-store.yaml#L54

test/uk/gov/hmrc/helptosavefrontend/repo/SessionStoreSpec.scala
sureshhmrc: yeah, this test some times runs upto 1 minute - don't want to run the unit tests long so commented out -

PR #740, opened 2018-11-12
HTS-1334: Moved Bars logic to BE

sureshhmrc:


app/uk/gov/hmrc/helptosavefrontend/services/HelpToSaveService.scala
```diff
@@ -125,6 +128,21 @@ class HelpToSaveServiceImpl @Inject() (helpToSaveConnector: HelpToSaveConnector)
...
+    EitherT(helpToSaveConnector.validateBankDetails(barsRequest).map[Either[String, Boolean]] { response ⇒
+      response.status match {
+        case Status.OK ⇒
+          Try((response.json \ "isValid").as[Boolean]) match {
+            case Success(isvalid) ⇒ Right(isvalid)
+            case Failure(error)   ⇒ Left(s"couldn't parse bars response from BE, error=${error.getMessage}. Body was ${maskNino(response.body)}")
+          }
+
+        case other ⇒
+          Left(s"Received unexpected status $other from BARS validation. Body was ${maskNino(response.body)}")
```
AndyHWChung: here and elsewhere in the FE I'd refrain from mentioning BARs and instead try to use something more generic like BankDetailsValidation or something - the BE should be the only thing concerned with how the bank details are validated, the FE doesn't need to know about BARs. That way, if we ever swap out BARs for something else, the FE code and logs would still make sense

app/uk/gov/hmrc/helptosavefrontend/controllers/BankAccountController.scala
```diff
@@ -77,31 +76,30 @@ class BankAccountController @Inject() (val helpToSaveService: HelpToSaveService,
...
+                    .fold(
+                      error ⇒ {
+                        logger.warn(s"Could not update session with bank details: $error")
+                        internalServerError()
+                      },
+                      _ ⇒ SeeOther(routes.RegisterController.getCreateAccountPage().url)
+                    )
+                } else {
+                  val formWithErrors = BankDetails.giveBankDetailsForm().fill(bankDetails)
+                    .withError("sortCode", BankDetailsValidation.ErrorMessages.sortCodeBarsInvalid)
```
AndyHWChung: we've still got a reference to BARS here and the line below

sureshhmrc: i thought these are specific to bars service -  let me take a closer look

sureshhmrc: fixed

AndyHWChung:
LGTM

PR #778, opened 2019-02-01
hts-1598 bank details error message updates

thebailers:


AndyHWChung:
there seems to be a broken unit test

PR #788, opened 2019-02-15
hts-1570 adds help and information page

thebailers:


app/uk/gov/hmrc/helptosavefrontend/views/helpinformation/how_sign_in.scala.html
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@()(implicit messages: play.api.i18n.Messages)
+
+<p class="bold">@Messages("hts.help-information.section.how-account-works.s2.p1")</p>
+<p>@Html(Messages("hts.help-information.section.how-account-works.s2.p2", "http://www.gov.uk/hmrc/app"))</p>
```
AndyHWChung: this URL seems to belong with the other URL's in application.conf (https://github.com/hmrc/help-to-save-frontend/blob/master/conf/application.conf#L225). If the URL needs changing in the future it'd be easier if it was read from application.conf. The same principle applies to the youtube embed codes - it'd be easier change both of them if we need to in the future if they sat in application.conf

thebailers: Have added these URLs to the app config.

app/uk/gov/hmrc/helptosavefrontend/controllers/IntroductionController.scala
```diff
@@ -66,4 +66,8 @@ class IntroductionController @Inject() (val authConnector: AuthConnector,
   def showPrivacyPage: Action[AnyContent] = unprotected { implicit request ⇒ implicit htsContext ⇒
     Ok(views.html.core.privacy())
   }
+
+  def getAboutPage: Action[AnyContent] = unprotected { implicit request ⇒ implicit htsContect ⇒
```
AndyHWChung: is this page only supposed to be accessible to people who have an account? If that were the case we'd want to make sure someone actually has an account here before displaying the content rather than use `unprotected` here which opens up the content to everyone

AndyHWChung: if we need to change this from `unprotected` this would be a BE change

conf/messages
```diff
@@ -685,3 +680,129 @@ global.error.InternalServerError500.heading=Sorry, we’re experiencing technica
...
+hts.help-information.section.bonuses.s1.p1=You can earn 2 tax-free bonuses.
+hts.help-information.section.bonuses.s1.h1=First bonus:
+hts.help-information.section.bonuses.s1.ul1-li1=earn up to £600
+hts.help-information.section.bonuses.s1.ul1-li2=paid after 2 years
+hts.help-information.section.bonuses.s1.h2=Final bonus:
+hts.help-information.section.bonuses.s1.ul2-li1=earn up to £600
+hts.help-information.section.bonuses.s1.ul2-li2=paid after 4 years
+hts.help-information.section.bonuses.s1.p2=Your bonus is calculated using your highest balance.
+hts.help-information.section.bonuses.s1.p3=You can withdraw money and still get any bonus money you’ve earned. You can earn £1,200 in bonuses in total.
+hts.help-information.section.bonuses.s1.p4=If you close your account before it matures after 4 years, you will not get a bonus.
```
Jackie1988: But don't they get a bonus after 2 years?

PR #803, opened 2019-05-20
DLS-245: Added Real user's account no to setup_regular_payment view

Jackie1988:


app/uk/gov/hmrc/helptosavefrontend/models/account/AccountNumber.scala
```diff
@@ -18,7 +18,7 @@ package uk.gov.hmrc.helptosavefrontend.models.account
 
 import play.api.libs.json.{Format, Json}
 
-case class AccountNumber(accountNumber: String)
+case class AccountNumber(accountNumber: Option[String])
```
lutnos: Now that AccountNumber holds an option of a String, is there a good reason for uk.gov.hmrc.helptosavefrontend.services.HelpToSaveServiceImpl.SubmissionSuccess to contain an Option[AccountNumber], rather than just an AccountNumber ?

Jackie1988: Doing that messes up the createAccount method in RegisterController, and I'm not sure how to re-work it

lutnos: Having made:

case class SubmissionSuccess(accountNumber: AccountNumber) extends SubmissionResult

and made helpToSaveService.createAccount return SubmissionSuccess(AccountNumber(None)) instead of SubmissionSuccess(None)

then in RegisterController.createAccount you could just _.accountNumber.fold in place of _.fold( at line 150

I think

Jackie1988: traverse is no longer available on line 128

lutnos: How about as in my https://github.com/hmrc/help-to-save-frontend/tree/dave/DLS-245/incorrect-payment-ref

Jackie1988: I dont think that will compile

PR #829, opened 2019-12-11
DLS-513 Button Focus Colour (Accessibility)

jordanrowe:


app/assets/stylesheets/help-to-save-frontend.scss
```diff
@@ -21,6 +21,21 @@
     outline: 3px solid #ffbf47;
 }
 
+.button:focus {
```
callum-gibbons: This will still apply to the 'button--link' buttons. Might be worth checking it looks ok still. If not, you can use :not(.button--link), to stop this style from applying.

jordanrowe: Looks to be okay
HTS Accessibility

PR #843, opened 2020-02-17
DLS-1854 for comprehension replaced with multiple folds and flatMap

rajanimohan22:
'for' comprehension replaced with multiple folds and flatMap

### DIFF
--- a/app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
+++ b/app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
@@ -94,8 +94,10 @@ class ReminderController @Inject() (val helpToSaveReminderService: HelpToSaveRem
                 },
                 emailRetrieved ⇒
                   emailRetrieved match {
-                    case Some(email) if !email.isEmpty ⇒
-                      helpToSaveReminderService.updateHtsUser(HtsUser(Nino(htsContext.nino), email, userInfo.forename, true, daysToReceive = DateToDaysMapper.d2dMapper.getOrElse(success.reminderFrequency, Seq(1))))
+                    case Some(email) if !email.isEmpty ⇒ {
+
+                      val htsUserToBeUpdated = HtsUser(Nino(htsContext.nino), email, userInfo.forename, true, daysToReceive = DateToDaysMapper.d2dMapper.getOrElse(success.reminderFrequency, Seq()))
+                      helpToSaveReminderService.updateHtsUser(htsUserToBeUpdated)
                         .fold(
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
```diff
@@ -80,44 +79,42 @@ class ReminderController @Inject() (val helpToSaveReminderService: HelpToSaveRem
...
-              })
-          }
-
-        }
-    )
+                },
+                emailRetrieved ⇒
+                  emailRetrieved match {
+                    case Some(email) ⇒
+                      if (email.length > 0)
```
lutnos: You can do this more tidily using a guard, so say:

      case Some(email) if !email.isEmpty ⇒
 
and later

      case Some(_) =>
                      logger.warn(s"Empty email retrieved for user: ${userInfo.nino}")

app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
```diff
@@ -80,44 +79,42 @@ class ReminderController @Inject() (val helpToSaveReminderService: HelpToSaveRem
...
+                        helpToSaveReminderService.updateHtsUser(HtsUser(Nino(htsContext.nino), email, userInfo.forename, true, daysToReceive = DateToDaysMapper.d2dMapper.getOrElse(success.reminderFrequency, Seq(1))))
```
lutnos: This line is too long, my personal preference would be to introduce a val for daysToReceive parameter

lutnos: That still comes out at more than 160 character for the line, I actually meant something like:

                      val daysToRetrieve = DateToDaysMapper.d2dMapper.getOrElse(success.reminderFrequency, Seq())
                      helpToSaveReminderService.updateHtsUser(HtsUser(Nino(htsContext.nino), email, userInfo.forename, true, daysToRetrieve))

app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
lutnos: Great, thanks

app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
rajanimohan22: Ok, let me do this

PR #851, opened 2020-03-02
DLS-2065 Remember users selection on the settings page

GujjalapudiA:


GujjalapudiA:
As per @chanelle comments,  this change is not required and it would go against HMRC design patterns to "pre select" an option, so closing this PR

PR #852, opened 2020-03-03
DLS-2054 Create front end for HTS reminders- Change email journey

GujjalapudiA:
change implemented


GujjalapudiA:
Hi Dave, please hold review, as per chanelle this page is already in place in live and we just need to add reminder link in page, I will confirm once it check in QA

lutnos:
OK, NP

PR #859, opened 2020-03-12
DLS-2129 HTS Reminders - Create test only end point 

GujjalapudiA:
Create test only end point  to get the reminder schedule details given NINO


lutnos:
This work is not required, the badly named help-to-save-reminder getifhtsuserexists endpoints already returns all of the data required

PR #898, opened 2020-05-28
DLS-2383 Add a scheduled help-to-save-frontend shuttering facility

GujjalapudiA:


app/uk/gov/hmrc/helptosavefrontend/auth/HelpToSaveAuth.scala
```diff
@@ -120,6 +123,14 @@ trait HelpToSaveAuth extends AuthorisedFunctions with AuthRedirects with Logging
       }
     }
 
+  private def isInMaintaince(currantTime: Long, maintainceTimes: Seq[appConfig.MaintainceTimes]): Boolean = {
```
lutnos, line 126: Spelling, this may have been Colm, should be currentTime, not when we eat raisins :-)

GujjalapudiA:
Ok Dave, could you please let me know the text to be displayed along with timings? do we need to ask content  team?

app/uk/gov/hmrc/helptosavefrontend/auth/HelpToSaveAuth.scala
```diff
@@ -120,6 +122,21 @@ trait HelpToSaveAuth extends AuthorisedFunctions with AuthRedirects with Logging
       }
     }
 
+  private def isInMaintenance(currentTime: Long, maintenanceTimes: Seq[appConfig.MaintainceTimes]): Boolean = {
+    val currentDateTime =
+      if (isSummerTime(LocalDateTime.now())) LocalDateTime.now().minusHours(1) else LocalDateTime.now()
+    val checkTimes = maintenanceTimes.map(
+      mt ⇒
+        if (currentDateTime.isAfter(mt.startTime) && currentDateTime.isBefore(mt.endTime)) {
+          endTime = mt.endTime.toString()
```
lutnos, line 131: No, you should not use a var to do that!

Why not have isMaintenance instead be a method that passes out an option of an end time, then you can pass that as the parameter of the MaintenanceException?

GujjalapudiA:
code has been refactored and committed

PR #903, opened 2020-07-17
DLS-2731 HTS - Fix ZAP issue

GujjalapudiA:
HTS - Fix ZAP issue


GujjalapudiA:
Thanks Dave, please let me know your suggestion for proper solution once you are ok with advices then I will continue with changes

lutnos:
[Slack community-security]https://hmrcdigital.slack.com/archives/C0JUC5VGS/p1595245331180800)

lutnos:
I now think this is why @JackieDev used base64 here: uk/gov/hmrc/helptosavefrontend/util/EmailVerificationParams.scala:24, to hide the same vulnerability.  I had always wondered about it, but not previously come across this explanation

lutnos:
Thanks @GujjalapudiA , we now know this code change is not needed, because there is no vulnerability there.  The AES GCM cypher we use is not vulnerable to a padding oracle attack.  @maruthi-thammineni will disable that test or alert in this ZAP .  Further to that, I now know why @JackieDev used a similar base64 three years ago, for which I could see no good reason.  Now I can annotate that with an explanation.

Thanks for your work on this

PR #915, opened 2020-10-14
DLS-3277 Safe Journey 

gavinharris43:


app/uk/gov/hmrc/helptosavefrontend/views/helpers/summary_row.scala.html
```diff
@@ -14,9 +14,13 @@
  * limitations under the License.
  *@
 
-@(label: String, value: String, change: Html)
+@(label: String, value: String, change: Html, hidden: Boolean = false)
```
lutnos, line 17: Can you explain to me how this hidden works, please?  That value will be set in execution of Play, so if you were expecting Optimizely to change it, well it can't, it can only change things in the DOM

lutnos (changes requested):
I don't think your hidden row will work

app/uk/gov/hmrc/helptosavefrontend/views/helpers/summary_row.scala.html
gavinharris43: hidden will apply this class to the div 
`.reminders-default{
     display: none;
 }`
witch will make it not visible but removal/change of this class will make the content visable

lutnos: I see, cheers

lutnos (approved):

PR #918, opened 2020-11-06
DLS-3462 Welsh changes

gavinharris43:


conf/messages
```diff
@@ -804,5 +804,21 @@ accessibility.statement.heading.seven = Technical information about this service
...
+hts.month.may = May
+hts.month.june = June
+hts.month.july = July
+hts.month.august  = August
+hts.month.september = September
+hts.month.october = October
+hts.month.november = November
+hts.month.december = December
+
+
```
lutnos, line 822: I'm confused, why are these here?

lutnos (changes requested):
We should not have to have our own versions of month names in our English messages file

conf/messages
gavinharris43: Removed

gavinharris43: Removed.

lutnos (approved):
Great, thanks

lutnos:
Release requires deployment of help-to-save-frontend, blocked on  DLS-3252 Implement welsh accessibility statement file into HTS

PR #920, opened 2020-11-24
DLS-3611: [RP] Use new head element for GTM and update play-ui.

rambabu-posa:


lutnos (approved):

rambabu-posa:
Upgraded govuk-template from 5.56.0-play-26 to 5.58.0-play-26

PR #921, opened 2020-12-01
DLS-1946 added handeling for expected 404 form reminders

gavinharris43:


lutnos (approved):
Fine, but I would encourage using play.mvc.Http.Status, e.g. OK or NOT_FOUND, rather than integer literals.  Don't go boiling the ocean though!

lutnos (approved):

PR #926, opened 2020-12-14
DLS-3256a: dependency upgrades and compilation warning reduction

LGoldfinch2:


LGoldfinch2:
For some reason, we are getting compilation warnings because the compiler is conflating the licensing at the top of the file for an unused import.

lutnos (approved):

PR #947, opened 2021-07-12
DLS-4193 Replacing whitelist occurances with allowlist 

rajanimohan22:
Replacing whitelist with allowlist


gavinharris43 (approved):
LGTM

PR #959, opened 2021-11-18
[SM][DLS-4667]removed no longer used page after_four_years and also r…

SeanMurray1:
…emoved related content. Also updated after 4 years so that the next month is shown.

Signed-off-by: Sean Murray <sean.murray@digital.hmrc.gov.uk>


app/uk/gov/hmrc/helptosavefrontend/views/helpinformation/after_4_years.scala.html
```diff
@@ -18,10 +18,17 @@
 
 @(maybeDate:Option[LocalDate])(implicit messages: play.api.i18n.Messages)
 
+@getNextMonth(date: LocalDate) = @{
+date.getMonthValue + 1 match {
```
lutnos, line 22: BAD! There should be better ways of doing that.  You should be able to add one month to a LocalDate and then get the month

lutnos (changes requested):

app/uk/gov/hmrc/helptosavefrontend/views/helpinformation/after_4_years.scala.html
SeanMurray1: Thanks for catching this Dave, I completely forgot that plusMonths() existed. I am making the changes now.

lutnos (approved):

PR #963, opened 2021-12-06
[SM][DLS-4640]Added closing date to reminder.

SeanMurray1:
Signed-off-by: Sean Murray <sean.murray@digital.hmrc.gov.uk>

### DIFF
--- a/app/uk/gov/hmrc/helptosavefrontend/models/HTSEvent.scala
+++ b/app/uk/gov/hmrc/helptosavefrontend/models/HTSEvent.scala
@@ -16,6 +16,7 @@
 
 package uk.gov.hmrc.helptosavefrontend.models
 
+import java.time.LocalDate
 import play.api.libs.json.{Format, JsValue, Json}
 import uk.gov.hmrc.helptosavefrontend.config.FrontendAppConfig
 import uk.gov.hmrc.helptosavefrontend.util.NINO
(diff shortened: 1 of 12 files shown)


app/uk/gov/hmrc/helptosavefrontend/models/HTSEvent.scala
```diff
@@ -82,7 +83,13 @@ object SuspiciousActivity {
...
 }
 
-case class HTSReminderAccount(nino : String, emailAddress: String, firstName: String, lastName: String, optInStatus: Boolean, daysToReceive: Seq[Int])
+case class HTSReminderAccount(nino : String,
+                              emailAddress: String,
+                              firstName: String,
+                              lastName: String,
+                              optInStatus: Boolean,
+                              daysToReceive: Seq[Int],
+                              accountClosingDate: Option[LocalDate])
```
lutnos, line 92: But accountClosingDate was since renamed to endDate

lutnos (changes requested):
But a lot of the change looks to be in the reminders controller.

The story asks for endDate to be set when the users accesses their account, not just when they try to look at their reminder schedule.

Or is my quick scan of the coding wrong?

lutnos (approved):

lutnos (changes requested):
I have no idea why any of these new frontend d endpoints or redirects would be needed.  We are simply updating a field on a backend (the reminders backend) based on data from another backend (help-to-save-proxy, perhaps via the help-to-save backend), sometimes when the frontend is called.

PR #966, opened 2022-01-04
DLS-4640

ayub96:


app/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountController.scala
```diff
@@ -72,6 +73,21 @@ class AccessAccountController @Inject() (
             e => {
               logger.warn(s"error retrieving Account details from NS&I, error = $e", htsContext.nino)
             }, {account =>
+              val accountEndDate: Future[Option[LocalDate]] =
```
lutnos, line 76: Sorry, but I think this behaves badly in a number of ways.  It is difficulty to test realiably, because the update occurs asynchronously to redirect, so there is no way of being sure when it has completed other than by the logging.  Also call for isClosed occurs asynchronously too, so there is no guarantee of the ordering and so of the affect.

There pattern of code already present for isClosed was not very good, I think.  By then following it the problems have been compounded.

lutnos (changes requested):
I will put together a couple of branches to show what the problems are here, in case it is not clear from my comment on the controller

app/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountController.scala
ayub96: It did feel odd that the redirect was not connected to any logic and that the connector results were mapped to loggers, but yes, I followed the format of what was already there. I understand the issue now.

nubz:
this is so out of date it will never merge now we've upgraded the front end... so closing this to avoid mishaps but will leave branch intact should there ever be a need to revisit it.

PR #996, opened 2023-01-03
update copyright

ginnyamazed:


ruthdas03:
The entry of the below package needs to be removed from AppDependencies to comply with bobby rule.
hmrc                     %% "reactivemongo-test"          % "5.0.0-play-28"     % "test",

A separate PR is raised to resolve the above and it's dependent classes.

PR #997, opened 2023-01-04
DLS-5395: [RP] Fixed bobby rule issue and remove MongoSupport.

rambabu-posa:


ruthdas03 (changes requested):
Looks good, just one minor change. I think we can remove  import play.core.PlayVersion , as it's not used.

ruthdas03 (approved):

PR #999, opened 2023-01-20
[DLS-5177] select-reminders and bank-account-page

ruthdas03:


app/uk/gov/hmrc/helptosavefrontend/views/core/confirm_check_eligibility.scala.html
```diff
@@ -18,25 +18,24 @@
...
+@pageTitle = @{
   Messages("hts.no-account.title.h1")
 }
 
-@main(
-  title = title
-) {
-  <h1 class="heading-large">@title</h1>
+@layout(pageTitle) {
+  <h1 class="heading-large">@pageTitle</h1>
```
nubz, line 33: this should be using classname `govuk-heading--l`

ruthdas03: Sure, I hope I can change <h1 class="heading-large"> to use classname .govuk-heading--l, in all these pages.

nubz:
we have agreed to close this in favour of recreating new requests from branch DLS-5176

PR #1000, opened 2023-01-23
DLS-5177 select reminders bank account page

ruthdas03:
merging to the branch DLS-5176


nubz (approved):
LGTM - we still need to find all other elements that are using old style class names, but merging now will let us split that work up more effectively

PR #1002, opened 2023-01-25
[DLS-5177]-createAccount-accountCreated-page

ruthdas03:


app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -48,51 +51,73 @@
...
       <strong>@accountNumber</strong>
-    </span>
+    </div>
   </div>
 
   <p>@Html(Messages("hts.register.create_account.confirmation-info.p1")) @email@Html(Messages("hts.register.create_account.confirmation-info.p2"))</p>
 
-<div class="form-group @if(hasErrors){form-group-error}">
-  <h2 class="heading-medium">@Messages("hts.register.create_account.what-you-need-to-do.title")</h2>
+<div class="govuk-form-group @if(hasErrors){govuk-form-group--error}">
```
nubz, line 73: this `form-group` wrapper is not needed when using components

app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -48,51 +51,73 @@
...
+  <h2 class="govuk-heading--m">@Messages("hts.register.create_account.what-you-need-to-do.title")</h2>
```
nubz, line 74: should be `govuk-heading-m` (using only a single dash)

app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -48,51 +51,73 @@
...
     <p>@Messages("hts.register.create_account.what-you-need-to-do.p1")</p>
```
nubz, line 75: all paragraphs should, by default, use the class `govuk-body` this ensures spacing and line heights etc are consistent

app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -48,51 +51,73 @@
...
+                content = Text(Messages("generic.no")),
+                value = Some("no"),
+                checked = form.data.values.toList.contains("no")
+            )
+            )
+        ))
+
+
+    </div>
+    <div class="govuk-form--group">
```
nubz, line 110: this button does not need to be wrapped in an HTML div

app/uk/gov/hmrc/helptosavefrontend/views/register/create_account.scala.html
```diff
@@ -40,119 +42,159 @@
...
+
     @if(bankDetails.rollNumber.nonEmpty) {
-      @summary_row(Messages("hts.register.create_account.your-bank-details.roll-number"), bankDetails.rollNumber.getOrElse(""),
-        summary_section_change(
-          (routes.RegisterController.changeBankDetails.url + "#rollNumber"),
-          Messages("hts.register.create_account.change-your-roll-number"),
-          "change-your-roll-number"
-        )
-      )
+      @viewHelpers.govukSummaryList(SummaryList(
```
nubz, line 158: this is not supposed to be a separate govukSummaryList but instead it should be a SummaryListRow inside the one and only summary list you already rendered above

app/uk/gov/hmrc/helptosavefrontend/views/register/create_account.scala.html
```diff
@@ -40,119 +42,159 @@
...
     }
-    @summary_row(Messages("hts.register.create_account.your-bank-details.account-name"), bankDetails.accountName,
-      summary_section_change(
-        (routes.RegisterController.changeBankDetails.url + "#accountName"),
-        Messages("hts.register.create_account.change-your-account-name"),
-        "change-your-account-name"
-      )
-    )
+
+      @viewHelpers.govukSummaryList(SummaryList(
```
nubz, line 171: same as above, this should only be a row and not an entirely new summary list

app/uk/gov/hmrc/helptosavefrontend/views/register/create_account.scala.html
```diff
@@ -40,119 +42,159 @@
...
+      value = Value(content = Text(Messages(display2CharFormat(bankDetails.sortCode.toString())))),
+      actions = Some(Actions(
+      items = Seq(ActionItem(
+      href = routes.RegisterController.changeBankDetails.url + "#sortCode",
+      content = Text(Messages("hts.register.create_account.change"))
+      ))
+      ))
+      ))))
+
+      @viewHelpers.govukSummaryList(SummaryList(
```
nubz, line 145: should not be a new summary list for each row! the summary list above should have a Sequence (which is a comma separated list) of SummaryListRows, in order to build the list given there are conditions for including some rows it might be better to have a separate method to build the rows and then feed in that as a single value

app/uk/gov/hmrc/helptosavefrontend/views/register/account_created.scala.html
```diff
@@ -46,69 +47,85 @@
...
+                form.errors.map { error =>
+                ErrorLink(
+                href = Some(s"#${error.key}"),
+                content = Text(messages("hts.payInNow.error"))
+                )
+                },
+                title = Text(messages("hts.global.error-summary.title"))
+            ))
+    }
+  <div class="govuk-panel govuk-panel--confirmation">
```
nubz, line 61: this could be using GovukPanel with something like 
```
@viewHelpers.govukPanel(Panel(
    title = Text(title),
    content = HtmlContent(messages("hts.account-created.title.p1"))
  ))
```

test/uk/gov/hmrc/helptosavefrontend/controllers/RegisterControllerSpec.scala
```diff
@@ -617,9 +617,9 @@ class RegisterControllerSpec
 
         val result = getAccountCreatedPage()
         status(result) shouldBe OK
-        contentAsString(result) should include("Account created")
+        contentAsString(result) should include("Continue")
```
nubz, line 620: I am not sure this is the best check that the correct page has rendered, this change only asserts a continue button exists somewhere on the page - whereas the previous check was for the h1  content - if we match that instead then this assures us the correct page renders.

nubz (approved):
LGTM

PR #1009, opened 2023-02-06
[DLS-6854] Refined Inconsistencies in Heading Sizes

ruthdas03:


nubz (changes requested):
we cannot update pages that are using main_template with govuk classes only - they must continue to use `heading-large` for large H1s

nubz:
There is no need to change all the heading classes to govuk ones on templates that still use govuk_wrapper.

PR #1025, opened 2023-02-27
[DLS-6835] upgraded check_your_email and reminder_cancel_confirmation

ruthdas03:


app/uk/gov/hmrc/helptosavefrontend/views/email/accountholder/check_your_email.scala.html
```diff
@@ -18,52 +18,52 @@
...
-    }
-    
-    <p class="subsection--wide push--bottom"><a href=@{frontendAppConfig.nsiManageAccountUrl} class="button" role="button">@messages("hts.global.back-to-account")</a></p>
+@viewHelpers.form(uk.gov.hmrc.helptosavefrontend.controllers.routes.AccountHolderController.onSubmit) {
+<p class="govuk-body">@Messages("hts.email-verification.existing-user.check-your-email.resend-1")
 
-  
+    @viewHelpers.govukButton(
+    Button(
+    attributes = Map("id" -> "resend-verification"),
```
nubz, line 61: to reproduce what we had before upgrading this button should have an extra class added to make it secondary button instead of a green, call-to-action button e.g. `classes="govuk-button--secondary",`

app/uk/gov/hmrc/helptosavefrontend/views/email/accountholder/check_your_email.scala.html
```diff
@@ -18,52 +18,52 @@
...
+    content = Text(messages("hts.email-verification.existing-user.check-your-email.resend-button-text"))
+    )
+    )
+    @Messages("hts.email-verification.existing-user.check-your-email.resend-2")</p>
+}
+    
+    <p class="govuk-body"><a class="govuk-link" href=@{frontendAppConfig.nsiManageAccountUrl}>@messages("hts.global.back-to-account")</a></p>
```
nubz, line 68: previously this link was styled as a button, to reproduce we should use govukButton and pass an `href` value to generate a link styles as button.

app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_cancel_confirmation.scala.html
```diff
@@ -34,7 +34,7 @@
 
     <h2 class="govuk-heading-m">@Messages("hts.reminder-confirmation.h2")</h2>
 
-    <p class="govuk-body">@Messages("hts.reminder-cancel-confirmation.p1-1") <a class="govuk-link" href=@routes.ReminderController.getEmailsavingsReminders.url>
+    <p class="govuk-body">@Html(Messages("hts.reminder-cancel-confirmation.p1-1")) <a class="govuk-link" href=@routes.ReminderController.getEmailsavingsReminders.url>
```
nubz, line 37: we don't need to wrap this message in Html - when I look at the message value it is a plain string (with no html tags within) that says "You can" in English and "Gallwch" in Welsh - worth checking sometimes if we see it in the old code we are upgrading that Html parsing is required - unnecessary or blanket Html wrapping is a [security concern ](https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=TEC&postingDay=2023%2F1%2F23&title=XSS+Protection+in+play+frontends#XSSProtectioninplayfrontends-play.twirl.api.Html)so we are encouraged to use only where necessary and we can confirm the html is safe

app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_cancel_confirmation.scala.html
```diff
@@ -17,32 +17,27 @@
...
-    <div class="govuk-box-highlight">
-        <h1 class="heading-xlarge">@title</h1>
-        <span class="heading-large">
-            @Messages("hts.reminder-cancel-confirmation.title.p1")
-        </span>
-    </div>
-            <h2 class="heading-medium">@Messages("hts.reminder-confirmation.h2")</h2>
+    <p class="govuk-body">@Html(Messages("hts.reminder-cancel-confirmation.p1-1")) <a class="govuk-link" href=@routes.ReminderController.getEmailsavingsReminders.url>
+    @Html(Messages("hts.reminder-cancel-confirmation.li-signup"))</a>
+        @Html(Messages("hts.reminder-cancel-confirmation.p1-2"))</p>
```
nubz, line 39: similarly these message values are just plain strings and do not need Html parsing, the original file used Html parsing but it didn't need to either

nubz (approved):

PR #1041, opened 2023-09-04
[DLS-8613] moved copy to be part of the hint copy

ruthdas03:


app/uk/gov/hmrc/helptosavefrontend/views/email/select_email.scala.html
```diff
@@ -40,7 +40,8 @@
       classes = "govuk-label--s"
     ),
     hint = Some(Hint(
-      content = HtmlContent(messages("hts.select-email.enter-new-address-hint")))),
+      content = HtmlContent(messages("hts.select-email.enter-new-address-hint") +
+                "<br>"+s"${messages("hts.select-email.enter-new-address-body-text")}"))),
```
nubz, line 44: we don't need to use string interpolation here, the output of message() is a string so you can concat the string without the interpolation syntax: 
```suggestion
                "<br>"+messages("hts.select-email.enter-new-address-body-text")))),
```

nubz (approved):

PR #1049, opened 2023-09-29
DLS-6752: Addressing PII Leaks for frontend

karolina-zaborska:
- removed response.body logging
- upgraded sbt-auto-build

### DIFF
--- a/app/uk/gov/hmrc/helptosavefrontend/connectors/EmailVerificationConnector.scala
+++ b/app/uk/gov/hmrc/helptosavefrontend/connectors/EmailVerificationConnector.scala
@@ -131,7 +131,7 @@ class EmailVerificationConnectorImpl @Inject() (http: HttpClient, metrics: Metri
       case other =>
         logger.warn(
           s"Received unexpected status $other from email verification" +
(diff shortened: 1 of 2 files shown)


TomRafferty (approved):

app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
```diff
@@ -164,8 +164,8 @@ class ReminderController @Inject() (
           },
           success =>
             htsContext.userDetails match {
-              case Left(missingUserInfos) =>
-                logger.warn(s"Email was verified but missing some user info $missingUserInfos")
```
LAJW, line 168: ❓ These missingInfo logs sound like logging an error rather than user data - if we're logging "missing fields" I assume these are the fields that the user did not fill, so we probably can log it.

LAJW (changes requested):

app/uk/gov/hmrc/helptosavefrontend/controllers/ReminderController.scala
TomRafferty: good spot!

LAJW (approved):

PR #1072, opened 2024-02-16
DLS-9947 - nino and DoB link UPDATE

Timidigital:


app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_dashboard.scala.html
```diff
@@ -38,14 +38,14 @@ <h1 class="govuk-heading-l">@messages("hts.reminder-dashboard.h2")</h1>
           s"""${messages("hts.reminder-dashboard.t1-1")} ${PeriodUtils.getMessage(period)} ${messages("hts.reminder-dashboard.t1-2")}""".stripMargin,
           routes.ReminderController.getSelectedRendersPage,
           messages("hts.reminder-dashboard.t1"),
-          Some(messages("hts.reminder-dashboard-change-cancel"))
+          (messages("hts.reminder-dashboard-change-cancel"))
```
nubz, line 41: this doesn't look right - you have changed the API for this helper and here there are no named parameters instead because they are anonymous then the sequence of the params is going to be used and the change link (which should remain an option and therefore needs the `Some()` wrapper) is not in the right position - to update we should name these parameters here now you have changed the order in the helper signature

app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_dashboard.scala.html
```diff
@@ -38,14 +38,14 @@ <h1 class="govuk-heading-l">@messages("hts.reminder-dashboard.h2")</h1>
...
         ),
         summaryRowHelper.summaryListRow(
           messages("hts.reminder-dashboard.t2"),
           email,
           routes.AccountHolderController.getUpdateYourEmailAddress,
           messages("hts.reminder-dashboard.t2"),
-          Some(messages("hts.reminder-dashboard-change"))
+          (messages("hts.reminder-dashboard-change"))
```
nubz, line 48: same as previous call to this helper, the param order has changed but these params are in the wrong order when parameter names are missing - use parameter names e.g. `changeLabel = Some(messages("hts.reminder-dashboard-change-cancel"))`

nubz (changes requested):

app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_dashboard.scala.html
Timidigital: reverted in [887c29b](https://github.com/hmrc/help-to-save-frontend/pull/1072/commits/887c29b3776f4dd7390c1270956b721cb50b76b5)

app/uk/gov/hmrc/helptosavefrontend/views/reminder/reminder_dashboard.scala.html
Timidigital: reverted in [887c29b](https://github.com/hmrc/help-to-save-frontend/pull/1072/commits/887c29b3776f4dd7390c1270956b721cb50b76b5)

PR #1078, opened 2024-02-28
DLS-9606 - Bootstrap v8 + Play30 Migration

TomRafferty:


LAJW (commented):
We need to sort out the play filter situation. They're being used on externaltest.

LAJW (approved):
Agreed to turn off allow lists

PR #1080, opened 2024-03-08
DLS-9948 Session email address and reminders

lutnos:
Retain session email address and reminders, so that change from check your answers works accessibly the way that it should


LAJW:
Superseded by: https://github.com/hmrc/help-to-save-frontend/pull/1081

PR #1082, opened 2024-04-24
DLS-10302: Content changes for help and information page

karolina-zaborska:


conf/messages
```diff
@@ -560,6 +560,9 @@ report.a.problem.link=Is there anything wrong with this page?
 # Help/information
 hts.help.information.title.h1=Help and information
 
+hts.help.information.h2=Ask HMRC online
+hts.help.information.p1=For more information about the Help to Save scheme <a class="govuk-link" href="https://www.tax.service.gov.uk/ask-hmrc/chat/help-to-save" rel="noreferrer noopener" target="_blank">ask HMRC’s digital assistant</a>.
```
LAJW, line 564: ❓ Why is it set to open in a new tab (target="_blank")? It doesn't state this on the requirement. Also why the rel attributes?
🚫 If they have to stay, I'm not a fan of duplication of the URL, class and rel tags. I'd break this down into two messages, and construct HTML in the template.

LAJW (changes requested):

conf/messages
karolina-zaborska: target="_blank" stays as it is preferred by the UCD to open the digital assistant in a new tab. All other insights applied

LAJW (approved):

PR #1089, opened 2024-09-17
DLS-10416 Check your details at the start

johno1985:


mahesh-hmrc:
Quick check and noticed that we don't have test coverage added for the new controller `CheckYourDetailsController`. If we can add that. Will go through rest of the PR. Thanks

app/uk/gov/hmrc/helptosavefrontend/controllers/EligibilityCheckController.scala
```diff
@@ -184,7 +184,11 @@ class EligibilityCheckController @Inject() (
   def getMissingInfoPage: Action[AnyContent] =
     authorisedForHtsWithInfo { implicit request => implicit htsContext =>
       htsContext.userDetails.fold(
-        missingInfo => Ok(missingUserInfo(missingInfo.missingInfo)),
+        { missingInfo =>
+          // DLS-10416 - Check in Kibana to see whether we can remove the Date of Birth warning
+          logger.info("Missing user info: " + missingInfo.missingInfo.toString())
```
mahesh-hmrc, line 189: Same as above, would be ideal not to log all details, but the fact that DOB is missing

test/uk/gov/hmrc/helptosavefrontend/controllers/CheckYourDetailsControllerSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+        contentAsString(result).trim should include(
+          "Your details"
+        )
+
+      }
+    }
+  }
+
+  private def doCommonChecks(doRequest: () => Future[Result]): Unit = {
+//    "redirect user to eligibility checks if there is no session found" in {
```
mahesh-hmrc, line 70: Is this still in works? Call to this is made in above case. if not required, can remove

app/uk/gov/hmrc/helptosavefrontend/controllers/CheckYourDetailsController.scala
```diff
@@ -0,0 +1,67 @@
...
+  ec: ExecutionContext
+) extends BaseController(cpd, mcc, errorHandler, maintenanceSchedule) with HelpToSaveAuth with EnrolmentCheckBehaviour
+    with Logging {
+
+  def checkYourDetails: Action[AnyContent] =
+    authorisedForHtsWithInfo { implicit request => implicit htsContext =>
+      checkIfAlreadyEnrolled { () =>
+        htsContext.userDetails match {
+          case Left(missingUserInfos) =>
+            logger.warn(s"user data missing: $missingUserInfos")
```
mahesh-hmrc, line 55: Probably ideal not to log the whole user info here. Has full name, DOB and contact details.
I understand this is to track the volume of users missing the DOB field, mentioned in DLS-10416.

However, if it's for that we are logging, better would be to just log the fact DOB is missing than the actual details.

test/uk/gov/hmrc/helptosavefrontend/controllers/CheckYourDetailsControllerSpec.scala
johno1985: Yep, still WIP, apologies, I should have marked it as such

app/uk/gov/hmrc/helptosavefrontend/controllers/CheckYourDetailsController.scala
johno1985: I believe this only logs field names, not the actual values

mahesh-hmrc: Oh wow, just had a look, it's a separate case class altogether for this, thought it was coming off userinfo. pretty sure that could be lot more simplified by data validation (similar to cats data) on the original userinfo model and log the fields missing

johno1985: Probably, but I didn't want to go and re-write everything that Lukasz did

app/uk/gov/hmrc/helptosavefrontend/controllers/EligibilityCheckController.scala
mahesh-hmrc: Discussion in the other message, nothing to address now

mahesh-hmrc (approved):

PR #1090, opened 2024-09-20
DLS-10958 | Add personal tax enrolment check as part of nino authorisation

mahesh-hmrc:
PR to address a security incident that has compromised some accounts. To address that, as part of HTS authentication
- Validate that the user session nino has a corresponding personal tax enrolment
- Validate that the user session nino matches to the nino from the personal tax enrolment

If any of the above validation steps fail, redirect the user to the personal tax endpoint.

### DIFF
--- a/test/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountControllerSpec.scala
+++ b/test/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountControllerSpec.scala
@@ -22,6 +22,7 @@ import play.api.test.FakeRequest
 import play.api.test.Helpers._
 import uk.gov.hmrc.auth.core.authorise.EmptyPredicate
 import uk.gov.hmrc.auth.core.retrieve.EmptyRetrieval
+import uk.gov.hmrc.auth.core.syntax.retrieved.authSyntaxForRetrieved
 import uk.gov.hmrc.domain.Nino
 import uk.gov.hmrc.helptosavefrontend.models.HtsAuth.AuthWithCL200
 import uk.gov.hmrc.helptosavefrontend.models.account.{Account, BonusTerm}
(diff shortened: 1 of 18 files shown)


test/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountControllerSpec.scala
```diff
@@ -135,8 +136,28 @@ class AccessAccountControllerSpec
         test(Some(HTSSession.empty))
       }
 
+      "redirect to protect tax if nino from user session doesn't have corresponding pt enrolment" in {
+        mockAuthWithNINORetrievalWithSuccess(AuthWithCL200)(mockedNINORetrieval and noPersonalTaxEnrolment)
+        mockSessionStoreGet(Right(Some(HTSSession.empty.copy(attemptedAccountHolderPageURL = Some("abc")))))
```
johno1985, line 141: I'm guessing that we probably don't need these two mocks as the request does not get that far

test/uk/gov/hmrc/helptosavefrontend/controllers/AccessAccountControllerSpec.scala
```diff
@@ -135,8 +136,28 @@ class AccessAccountControllerSpec
...
+        mockEnrolmentCheck()(Right(EnrolmentStatus.Enrolled(true)))
+
+        val result = doRequest()
+        status(result) shouldBe 303
+        redirectLocation(result) shouldBe Some(appConfig.protectTaxInfoUrl)
+      }
+
+      "redirect to protect tax if nino from user session doesn't match nano from PT enrolment" in {
+        mockAuthWithNINORetrievalWithSuccess(AuthWithCL200)(mockedNINORetrieval and enrolmentsWithNoMatchingNino)
+        mockSessionStoreGet(Right(Some(HTSSession.empty.copy(attemptedAccountHolderPageURL = Some("abc")))))
```
johno1985, line 151: I'm guessing that we probably don't need these two mocks as the request does not get that far

johno1985 (dismissed):
Would love to know how exactly it happened.

johno1985 (approved):

PR #1095, opened 2024-11-04
DLS-11031: Code changes for double click prevention.

vamsikrishnapujari:


ruthdas03:
There are html files where Button is missing the attribute to prevent double click

`email_updated.scala.html `
`accountholder/check_your_email.scala.html`

Are we not adding attribute `"data-prevent-double-click"->"true"`   for Button with href

vamsikrishnapujari:
Hi @ruthdas03 , I made changes for the above tow mentioned files as well. Please check it once and let me know.

ruthdas03 (approved):
ATs need to be run

PR #1116, opened 2026-05-18
DLS-13035|Welsh translation for DOB

gajalakshmiseetharaman:
AT's & PT's are working as expected

image
image


app/uk/gov/hmrc/helptosavefrontend/views/helpers/DateUtils.scala
```diff
@@ -36,4 +36,21 @@ object DateUtils {
     val monthValue: String = toLocalisedStringMonth(date)
     s"${date.format(dayExtractor)} $monthValue ${date.format(yearExtractor)}"
   }
+
```
ShivangiShah42, line 40: can we please add  unit test?

gajalakshmiseetharaman: Unit test case added

ShivangiShah42, line 40: Sorry could have adde this comment along with unit test, but can we also simplified the logic here? also ResolverStyle is not neccessory for formatting

ShivangiShah42 (approved):

PR #1118, opened 2026-05-27
DLS-13034: Committing code changes for addition of Notification banne…

vamsikrishnapujari:
…r for resend the email button.

### DIFF
--- a/app/uk/gov/hmrc/helptosavefrontend/controllers/AccountHolderController.scala
+++ b/app/uk/gov/hmrc/helptosavefrontend/controllers/AccountHolderController.scala
@@ -88,47 +88,80 @@ class AccountHolderController @Inject() (
       htsContext.firstName.fold[Future[Result]](
         SeeOther(routes.EmailController.confirmEmailErrorTryLater.url)
       ) { name =>
-        checkIfAlreadyEnrolled(
-          _ =>
-            UpdateEmailForm.verifyEmailForm
-              .bindFromRequest()
-              .fold(
-                formWithErrors => {
-                  BadRequest(updateEmailAddress(formWithErrors))
-                },
-                (details: UpdateEmail) =>
-                  emailValidation.validate(details.email).toEither match {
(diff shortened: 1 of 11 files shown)


ZaraFarzan (dismissed):
LGTM

app/uk/gov/hmrc/helptosavefrontend/views/email/accountholder/check_your_email.scala.html
```diff
@@ -54,16 +55,24 @@ <h2 class="govuk-heading-m">@messages("hts.email-verification.existing-user.chec
 @viewHelpers.form(uk.gov.hmrc.helptosavefrontend.controllers.routes.AccountHolderController.onSubmit) {
 <p class="govuk-body">@Messages("hts.email-verification.existing-user.check-your-email.resend-1")
```
mahesh-hmrc: Looks like we are not closing this `<p>` tag. Check pls

app/uk/gov/hmrc/helptosavefrontend/controllers/AccountHolderController.scala
```diff
@@ -127,8 +127,15 @@ class AccountHolderController @Inject() (
       }
     }(loginContinueURL = routes.AccountHolderController.onSubmit.url)
 
+  def resendEmail: Action[AnyContent] =
```
mahesh-hmrc: This flow seems wrong as it doesn't send verification before marking the session value `emailResent` as true. It's not inline with the  onSubmit action which is what sends email verification. Flow is slightly different to EmailController, so this needs to be handled correctly here.

Action [here](https://github.com/hmrc/help-to-save-frontend/pull/1118/changes#diff-7b0d21b753d6ef49d70d7420bfad15e3dba6fec7d9376844db3e45dae1bcc9d2R61) is straight away setting the session value but not triggering anything for sending email.

ZaraFarzan (approved):
LGTM

PR #1122, opened 2026-07-06
DLS-13169: Committing code changes for adding contextual copy for cha…

vamsikrishnapujari:
…nge links.


ShivangiShah42 (dismissed):

conf/messages
```diff
@@ -382,6 +382,8 @@ hts.register.create_account.change-your-account-number=Change your account numbe
 hts.register.create_account.change-your-roll-number=Change your roll number
 hts.register.create_account.change-your-account-name=Change your account name
 hts.register.create_account.address=Address
+hts.register.create_account.change-address-hidden-text=Address
```
ruthdas03: Need not create a new key value pair for Address we can use the existing one ts.register.create_account.address=Address

ruthdas03 (changes requested):

conf/messages
vamsikrishnapujari: Done, updated PR.

ruthdas03 (approved):

PR #1127, opened 2026-08-12
DLS-13312:[HTS] Update Title on Are you sure you want to close your account Page

ZaraFarzan:


vamsikrishnapujari (dismissed):
LGTM

app/uk/gov/hmrc/helptosavefrontend/views/closeaccount/close_account_are_you_sure.scala.html
```diff
@@ -65,7 +65,7 @@
 }
 
 @layout(
-    title = title + " " + Messages("hts.global.accountholder.title-suffix"),
```
gajalakshmiseetharaman, line 68: I couldn't see any usage of this key in the messages. If there are no dependencies on it, we could safely remove it from the message files as well.

gajalakshmiseetharaman (changes requested):

app/uk/gov/hmrc/helptosavefrontend/views/closeaccount/close_account_are_you_sure.scala.html
vamsikrishnapujari: Good catch, removed now.

gajalakshmiseetharaman (approved):

